PR #39, opened 2015-01-09
Add authors to front-page

trendelkampschroer:
I suggest to add a list of authors to the front-page of pyemma.


marscher:
I would not mind if you push such a little directly on devel :)

Issue #44, opened 2015-01-15
conda recipe only install scripts wrappers, but not their content


marscher:
has been reported to upstream conda/conda-build#296

marscher:
try to use setuptools entry_points for scripts and mimic them meta.yml

PR #54, opened 2015-01-28
Fix scripts with conda

marscher:
fix for #44


marscher:
release 1.0.2 and update conda recipe

PR #59, opened 2015-02-13
[utility][coordinates] added function to compute autocorrelation functio...

fabian-paul:
Hi Martin, I just moved the code from the old mm_acf into a new module and cleaned it up. 
Can you merge this please? Of course, you are free to move the new module (pyemma.coordinates.acf) somewhere else in the hierarchy if it fits better.
Fabian

### DIFF
--- a/pyemma/coordinates/acf.py
+++ b/pyemma/coordinates/acf.py
@@ -1,0 +1,105 @@
+import numpy as np
+import sys
+
+__author__ = 'Fabian Paul'
+__all__ = ['acf']
+
+def acf(trajs, stride=1, max_lag=None, subtract_mean=True, normalize=True, mean=None):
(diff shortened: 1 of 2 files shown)


marscher:
Thank you! This feature would actually nicely fit into the transformer logic, so we can process large data. Is it possible to calculate FFT chunkwise?

marscher:
have a look at transformers.tica to get an idea for the logic.

fabian-paul:
Yes, it should be possible to calculate the acf chunk-wise for lag times tau_max < chunk_size.
Besides that, there are many possible optimizations:
- calculate the FFT for one trajectory at a time
- calculate the FFT for one observable at a time
- subsample the trajectory with a stride to fit data with a longer lag time into one chunk.

marscher:
it then would be nice to have this in the transformer pipeline also. Should not be to hard to refactor.

PR #81, opened 2015-02-21
[coordinates] (maybe) bugfix in regspatial clustering

fabian-paul:


marscher:
makes no sense to repeat this step for every traj. I've fixed the deletion of the temporary already.

PR #85, opened 2015-02-24
Shortcut decorator to avoid cluttering of APIs

marscher:
new decorator "@shortcut" defined to create a function alias in the module, where the original function is defined. Docstrings are preserved and the alias is being added to the module (appended to __all__)


trendelkampschroer:
Cool!

PR #108, opened 2015-03-09
[msm] Fixed imports to avoid populating the msm namespace with obsolete ...

trendelkampschroer:
...names


marscher:
thanks. This PR also contains a bug fix for chapman kolmogorov test?

trendelkampschroer:
No this does only contain a fix in the imports in us and in the msm/api.py. The fix for the estimation and alanlysis api is not included. It does not contain a fix for the chapman-kolmogorov test either.

trendelkampschroer:
Maybe it is better to wait with merging until the tests for the coordinates modules are not failing anymore.

pyemma/msm/ui/chapman_kolmogorov.py
```diff
@@ -127,7 +131,22 @@ def chapman_kolmogorov(dtrajs, lag, K, nsets=2, sets=None):
         C=Ccc_k.toarray()
         for l in range(len(sets)):            
             w=np.zeros(len(lcc))
-            inds = lccmap.map(sets[l])
+            """There is a problem here - the initial sets will probably be bigger than the lcc at time k*\tau"""
```
marscher: this seems like a bugfix to me

marscher:
nevertheless all global imports in api modules are visible in IPython since the __all__ variable is ignored. I have prepared a branch to fix this by hiding global imports with underscores. There is no way around it.

trendelkampschroer:
This seems to be a bug in the ipython notebook app. In the 'normal' ipython console I cn only import the functions in __all__ plus the modules api, dense and sparse.

pyemma/msm/ui/chapman_kolmogorov.py
trendelkampschroer: No this is 'unfortunately' just a note that this needs a bug fix indicating where and why the problem occurs.

marscher: Am 09.03.2015 um 12:21 schrieb Benjamin Trendelkamp-Schroer:
> In pyemma/msm/ui/chapman_kolmogorov.py
> <https://github.com/markovmodel/PyEMMA/pull/108#discussion_r26027976>:
> 
>> @@ -127,7 +131,22 @@ def chapman_kolmogorov(dtrajs, lag, K, nsets=2, sets=None):
>>          C=Ccc_k.toarray()
>>          for l in range(len(sets)):            
>>              w=np.zeros(len(lcc))
>> -            inds = lccmap.map(sets[l])
>> +            """There is a problem here - the initial sets will probably be bigger than the lcc at time k*\tau"""
> 
> No this is 'unfortunately' just a note that this needs a bug fix
> indicating where and why the problem occurs.
> 
> —
> <https://github.com/markovmodel/PyEMMA/pull/108/files#r26027976>.
> 
I'm talking of the ten lines of code below this comment, dunno how to
comment on a range of lines.

trendelkampschroer:
This pull. requests implements  the changes in the msm high-level API and add unit test for the high-level API. It contains also some bug-fixes and the addition of the corresponding unit-tests in the msm.ui package.

PR #136, opened 2015-03-19
[tica] fixed calculation of covariance matrix. Introduced n_short attrib...

gph82:
...ute to count too short

trajectories in transformer. TODO: this attribute has to be reset somewhere


marscher:
this commit was included in #140

PR #153, opened 2015-03-27
[featurizer] impl hashing features to avoid to double add them and added test.

marscher:
These changes avoids that the user add the same feature (eg. same indexes for distance calculation twice). Even for the CustomFeature which gives the ability to call arbitrary functions this behaviour works.

I've also removed the index attributes of featurizer itself, since those are stored in the feature classes.

### DIFF
--- a/pyemma/coordinates/io/featurizer.py
+++ b/pyemma/coordinates/io/featurizer.py
@@ -1,5 +1,4 @@
-from pyemma.util.annotators import deprecated
-__author__ = 'noe'
+__author__ = 'noe, marscher'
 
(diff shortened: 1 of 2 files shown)


trendelkampschroer:
Cool, I definitely want to learn how you implemented hashs for ndarrays.

marscher:
I combined the hashes of shape, stride and underlying (readonly) buffer 
with xor. This is more robust and efficient, than calculating the hash 
of the string representation.

trendelkampschroer:
Godd to know.

PR #161, opened 2015-03-29
Improve featurization

franknoe:
Debugged, enhanced and refactored Featurizer:

* Removed various bugs and problems
* Wrote basic unit tests for almost all functions (pls add test for backbone dihedrals!)
* Degree option for dihedrals and angles
* Added general dihedral feature
* Added periodic flag for distance and contact calculations
* Added documentation to all functions
* dimension is now an attribute of feature, and has changed from a Featurizer method to an attribute


franknoe:
Stepping back from making Featurizer.dimension() a property. Causing too many problems.

Issue #152, opened 2015-03-27
angle/dihedral features vs averaging


marscher:
should we convert them via sin/cos?
Maybe add a flag to toggle the behaviour?

trendelkampschroer:
Hi,

this is a very important issue.

According to wikipedia radians is 'the standard unit' of angular data. So I think we should stick to it (in terms of units).

In terms of computations one has to distinguish between real-valued and circular data. Computation of many important quantities differ a lot, e.g. mean and covariances. If you are thinking about clustering then it is also important to discuss how to cluster angular data. 

It might be that a representation in terms of (cos, sin) paris does the trick in all relevant cases. I can refer to some literature if needed.

marscher:
the mean calculation is not well defined in angular space. Think of -179 
and 179, the mean should be +-180, but in our impl it would be 0!

So whenever a mean gets calculated (PCA, TICA, Kmeans) this will fail!

franknoe:
Ability to compute cos/sin would certainly be useful. I would say this 
is a different feature though.

-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

franknoe:
Am 27/03/15 um 21:17 schrieb Benjamin Trendelkamp-Schroer:
>
> According to wikipedia radians is 'the standard unit' of angular data. 
> So I think we should stick to it (in terms of units).
>
I would use a flag to switch between radians and degrees (but not 
between direct and cos/sin representation that I consider a different 
feature).

My feeling is that for MD people an angle is more commonly used than 
radians, but the default should probably be the one that is compatible 
with numpy operations on the data (probably radians). What do you think?



-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

gph82:
I agree with Frank: sin,cos should be another feature. I mean, the user has to be capable of computing angles. Think of our many Di-Ala MSMs, all done in angular space. We want to still be able to do those with PyEmma, right?

[...]

PR #162, opened 2015-03-29
Improve featurization

franknoe:
Debugged, enhanced and refactored Featurizer:

* Removed various bugs and problems
* Wrote basic unit tests for almost all functions (pls add test for backbone dihedrals!)
* Degree option for dihedrals and angles
* Added select atoms feature
* Added general dihedral feature
* Added periodic flag for distance and contact calculations
* Added documentation to all functions
* dimension is now an attribute of the Features and doesn't need to be set separately


marscher:
Am 29.03.2015 um 08:37 schrieb Frank Noe:
> Debugged, enhanced and refactored Featurizer:
>
>   * Removed various bugs and problems
>   * Wrote basic unit tests for almost all functions (pls add test for
>     backbone dihedrals!)
Thats really great! What's the molecular test system?  Why do we need a 
trajectory for testing the Featurizer (shouldn't a topology be enough?)
>   * Degree option for dihedrals and angles
nice, this is a partial fix for already #152?
>   * Added general dihedral feature
>   * Added periodic flag for distance and contact calculations
Maybe we should add a kwargs to all of the wrappers of mdtraj to support 
future developments and add more freedom to the user?
>   * Added documentation to all functions
>   * dimension is now an attribute of the Features and doesn't need to be
>     set separately
Makes more sense than setting this in Featurizer!
>
> ------------------------------------------------------------------------
>
>
>         You can view, comment on, or merge this pull request online at:
>
> https://github.com/markovmodel/PyEMMA/pull/162
>
>
>         Commit Summary
>
>   * doc changes in pcca
Why is that in this PR?
>   * Merge branch 'devel' of https://github.com/markovmodel/pyemma into devel
>   * Merge branch 'devel' of https://github.com/markovmodel/pyemma into devel
>   * Debugged and refactored featurizer
>   * Documented Featurizer
>   * coordinates.io.featurizer: Added Unit Tests and removed some bugs
>   * Featurizer: Added degree option to angles and dihedrals
>   * Featurizer: dimension is now provided by feature, and dimension is a
>     property
>   * Featurizer: removed iteration over set(active_features) because sets
>     are unordered which can lead to a random (unexpected) feature sequence
Is the order of the features really important? We can also use an 
OrderedSet here, if thats of matter.
>   * Added general dihedral feature to featurizer
>   * Stepping back: MDFeaturizer.dimension() is again a function.
>     Changing to property caused too many failures

pyemma/coordinates/io/featurizer.py
```diff
@@ -67,45 +67,158 @@ def __init__(self, func=None, *args, **kwargs):
     def describe(self):
         return "override me to get proper description!"
 
+    @property
+    def dimension(self):
+        # override me to get correct dimension
+        return 0
```
marscher: we should raise here, since this introduces an silent hard to debug failure!

pyemma/coordinates/io/featurizer.py
```diff
@@ -67,45 +67,158 @@ def __init__(self, func=None, *args, **kwargs):
...
+    def __init__(self, top, distance_indexes, periodic = True):
+        DistanceFeature.__init__(self, top, distance_indexes, periodic=periodic)
         self.prefix_label = "INVDIST:"
 
     def map(self, traj):
-        return 1.0 / mdtraj.compute_distances(traj, self.distance_indexes)
+        return 1.0 / mdtraj.compute_distances(traj, self.distance_indexes, periodic=self.periodic)
+
+
+class ContactFeature(DistanceFeature):
```
marscher: why not use http://mdtraj.org/latest/api/generated/mdtraj.compute_contacts.html#mdtraj.compute_contacts?

franknoe:
Am 29/03/15 um 14:56 schrieb Martin K. Scherer:
> > * Wrote basic unit tests for almost all functions (pls add test for
> > backbone dihedrals!)
> Thats really great! What's the molecular test system? Why do we need a
> trajectory for testing the Featurizer (shouldn't a topology be enough?)
It's 100 frames with Ca-atoms of BPTI from the DEShaw simulation.

The trajectory is very small. I could have used 10 frames for this test, 
but I would like to use it for other tests too (e.g. PCA).

A topology is not enough because a trajectory results in a multiple 
feature vectors. Moreover, I want to use the file for other tests.
> > * Degree option for dihedrals and angles
> nice, this is a partial fix for already #152?
> > * Added general dihedral feature
> > * Added periodic flag for distance and contact calculations
> Maybe we should add a kwargs to all of the wrappers of mdtraj to support
> future developments and add more freedom to the user?
>
I agree
> > * Featurizer: removed iteration over set(active_features) because sets
> > are unordered which can lead to a random (unexpected) feature sequence
> Is the order of the features really important? 
Definitely. If I add first distances and then angles, I want to know 
that I can interpret the resulting feature vector in the same way.

In addition, set(...) was done in map, which is called many times for 
data-streaming mode. Thus there is no guarantee that we will have even 
the same sequence while the data is streamed, leading to random 
reordering of the data.
> We can also use an
> OrderedSet here, if thats of matter.
Why do we need that? If you really want to exclude double adding of the 
same features we could simple if the feature exists in the list while 
adding. E.g. write an _append_feature(f) function which does checks like 
that if needed.
The number of different features will be small (typically 1-3), so 
there's no benefit from a hashable collection here.


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

pyemma/coordinates/io/featurizer.py
franknoe: Am 29/03/15 um 16:01 schrieb Martin K. Scherer:
>
> In pyemma/coordinates/io/featurizer.py 
> <https://github.com/markovmodel/PyEMMA/pull/162#discussion_r27354664>:
>
> > +class ContactFeature(DistanceFeature):
>
> why not use 
> http://mdtraj.org/latest/api/generated/mdtraj.compute_contacts.html#mdtraj.compute_contacts?
>
mdtraj.compute_contact seems to compute distances too rather than 
distances. At least the implementation I have found in the devel branch 
did not compute contacts. I have nothing against changing it if the 
behavior stays the same.


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
> Why do we need that? If you really want to exclude double adding of the
> same features we could simple if the feature exists in the list while
> adding. E.g. write an _append_feature(f) function which does checks like
> that if needed.
> The number of different features will be small (typically 1-3), so
> there's no benefit from a hashable collection here.
I agree, its not worth it for such a small collection.

I've merged my changes locally and will improve it to avoid double
features backed up by a list.

marscher:
the equality checks are still done by hashing, but backend is an ordered list again.

The select_backbone method still fails - added a test which just calls the method.

franknoe:
Am 29/03/15 um 17:49 schrieb Martin K. Scherer:
>
> The select_backbone method still fails - added a test which just calls 
> the method.
>
> —
> <https://github.com/markovmodel/PyEMMA/pull/162#issuecomment-87431123>.
>
This one should work:

traj.topology.select("backbone and (name C or name CA or name N)")


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
thanks that works.

PR #172, opened 2015-03-31
Modified Ben's cool MSM object to include LEVs and REVs if wanted

gph82:
Extended the MSM object a bit so that LEVs and REVs of a given MSM object do not live "outside" the object itself. Once you have an MSM object containing T, C, LargeConSet etc...it seems artificial not to have them there.

@trendelkampschroer: I used the "get_LEVs" convention for the first time they're computed. Perhaps there's a better way? Or perhaps we should calculate them directly? That might be very time consuming, since it might involve diagonalization of large matrices

### DIFF
--- a/pyemma/msm/ui/msm.py
+++ b/pyemma/msm/ui/msm.py
@@ -102,8 +102,9 @@ class MSM(object):
         self.mu = statdist(self.T)
 
         """ First left and right EVs"""
-        self.LEVs = eigenvectors(self.T, k = np.min([10, self.T.shape[0]-1]), right=False )          
-        self.REVs = eigenvectors(self.T, k = np.min([10, self.T.shape[0]-1]), right=True )
+        k = np.min([10, self.T.shape[0]-2])
+        self.LEVs = eigenvectors(self.T, k = k, right=False )          
+        self.REVs = eigenvectors(self.T, k = k, right=True )
 
         self.computed = True
 
--- a/pyemma/msm/ui/msm.py
+++ b/pyemma/msm/ui/msm.py
@@ -102,8 +102,8 @@ class MSM(object):
         self.mu = statdist(self.T)
 
         """ First left and right EVs"""
-        self.LEVs = eigenvectors(self.T, k = 10, right=False )          
-        self.REVs = eigenvectors(self.T, k = 10, right=True  )          
+        self.LEVs = eigenvectors(self.T, k = np.min([10, self.T.shape[0]-1]), right=False )          
+        self.REVs = eigenvectors(self.T, k = np.min([10, self.T.shape[0]-1]), right=True )
 
         self.computed = True
 
--- a/pyemma/msm/ui/msm.py
+++ b/pyemma/msm/ui/msm.py
@@ -11,7 +11,7 @@ __docformat__ = "restructuredtext en"
 import numpy as np
 
 from pyemma.msm.estimation import cmatrix, largest_connected_set, connected_cmatrix, tmatrix
-from pyemma.msm.analysis import statdist, timescales
+from pyemma.msm.analysis import statdist, timescales, eigenvectors
 
 __all__ = ['MSM']
 
@@ -67,11 +67,17 @@ class MSM(object):
         self.T = None
 
         """Stationary distribution"""
-        self.mu = None
+        self.mu = []
 
         self.computed = False
 
         self.estimate_on_lcc = estimate_on_lcc
+
+        """Left eigenvectors"""
+        self.LEVs = []
+
+        """Right eigenvectors"""
+        self.REVs = []
 
         if compute:
             self.compute()
@@ -128,10 +134,26 @@ class MSM(object):
     @property
     def stationary_distribution(self):
         self._assert_computed()
-        if self.mu is None:
+        if np.size(self.mu) == 0:
             self.mu = statdist(self.T)
         return self.mu
 
-    def get_timescales(self, k):
+    def get_timescales(self, k = 10):
         ts = timescales(self.T, k=k, tau=self.tau)
         return ts
+
+    def get_LEVs(self, k = 10):
+        self._assert_computed()
+        
+        if np.size(self.LEVs) == 0:
+           self.LEVs = eigenvectors(self.T.toarray(), k = k, right=False )
+        else:
+           print "LEVs already computed. Use MSM.LEVs"
+
+    def get_REVs(self, k = 10):
+        self._assert_computed()
+        
+        if np.size(self.REVs) == 0:
+           self.REVs = eigenvectors(self.T.toarray(), k = k, right=True  )
+        else:
+           print "REVs already computed. Use MSM.REVs"
--- a/pyemma/msm/ui/msm.py
+++ b/pyemma/msm/ui/msm.py
@@ -67,17 +67,17 @@ class MSM(object):
         self.T = None
 
         """Stationary distribution"""
-        self.mu = []
+        self.mu = None
 
         self.computed = False
 
         self.estimate_on_lcc = estimate_on_lcc
 
         """Left eigenvectors"""
-        self.LEVs = []
+        self.LEVs = None
 
         """Right eigenvectors"""
-        self.REVs = []
+        self.REVs = None
 
         if compute:
             self.compute()
@@ -134,7 +134,7 @@ class MSM(object):
     @property
     def stationary_distribution(self):
         self._assert_computed()
-        if np.size(self.mu) == 0:
+        if self.mu is None:
             self.mu = statdist(self.T)
         return self.mu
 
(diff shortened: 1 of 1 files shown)


franknoe:
Good. I am planning to extend the msm object significantly in the 
future, this is the right direction.

Am 31/03/15 um 16:31 schrieb gph82:
>
> Extended the MSM object a bit so that LEVs and REVs of a given MSM 
> object do not live "outside" the object itself. Once you have an MSM 
> object containing T, C, LargeConSet etc...it seems artificial not to 
> have them there.
>
> @trendelkampschroer <https://github.com/trendelkampschroer>: I used 
> the "get_LEVs" convention for the first time they're computed. Perhaps 
> there's a better way? Or perhaps we should calculate them directly? 
> That might be very time consuming, since it might involve 
> diagonalization of large matrices
>
> ------------------------------------------------------------------------
>
>
>         You can view, comment on, or merge this pull request online at:
>
> https://github.com/markovmodel/PyEMMA/pull/172
>
>
>         Commit Summary
>
>   * Modified Ben's cool MSM object to include LEVs and REVs if wanted
>
>
>         File Changes
>
>   * *M* pyemma/msm/ui/msm.py
>     <https://github.com/markovmodel/PyEMMA/pull/172/files#diff-0> (30)
>
>
>         Patch Links:
>
>   * https://github.com/markovmodel/PyEMMA/pull/172.patch
>   * https://github.com/markovmodel/PyEMMA/pull/172.diff
>
> —
> <https://github.com/markovmodel/PyEMMA/pull/172>.
>


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

trendelkampschroer:
Hi Guillermo,

thanks a lot for your effort. Can you please state the reason why you use empty lists instead of None to initialize the attributes that have not been computed?

trendelkampschroer:
Hi Guillermo,

it is also 'dangerous' to just set k=10, because the code will be broken for all MSMs with n<10. I sugest to use k as a mandatory keyword. The user has to specify how many eigenvalues and time-scales he wants.

gph82:
I thought that element wise comparison to None would fail (I used to get these notificiations, perhaps due to outdated np version?). You can switch it back

gph82:
As for k, perhaps take k = np.min([10, len(T)])?

trendelkampschroer:
Hi, 
 it is fine to do

x = np.arange(5)
x is None 
-> False

gph82:
yeah, I noticed. Switch it back, then. Thanks!

On 03/31/2015 05:44 PM, Benjamin Trendelkamp-Schroer wrote:
> Hi,
> it is fine to do
>
> x = np.arange(5)
> x is None
> -> False
>
> —
> <https://github.com/markovmodel/PyEMMA/pull/172#issuecomment-88138368>.
>

-- 
Dr. Guillermo Pérez-Hernández
Freie Universität Berlin
Institute for Mathematics
Arnimallee 6
D-14195 Berlin
tel       0049 30 838 75776

trendelkampschroer:
@gph82  Please close this pull-request and resubmit the discussed changes

franknoe:
I think k should be optional, by default the object should operate on 
dense matrices where all eigenvalues are computed and available. I 
suggest the default should be to compute all available eigenvalues, 
unless the user sets a smaller value. And even then if they set n=10 and 
the matrix is only 3x3, a warning should be raised and only 3 should be 
computed.


Am 31/03/15 um 17:12 schrieb Benjamin Trendelkamp-Schroer:
>
> Hi Guillermo,
>
> it is also 'dangerous' to just set k=10, because the code will be 
> broken for all MSMs with n<10. I sugest to use k as a mandatory 
> keyword. The user has to specify how many eigenvalues and time-scales 
> he wants.
>
> —
> <https://github.com/markovmodel/PyEMMA/pull/172#issuecomment-88127560>.
>


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
I just rewrote so that it does what Frank says. 

The "Fallunterscheidung" is a bit ugly, perhaps there is a cleaner way of writing it. 

One question if we want *all* eigenvectors, there is no point in using a sparse diagonalization routine, correct? 

That is why there are four cases.

trendelkampschroer:
Thanks Guillermo, you are correct. It makes no sense to use sparse types and have methods that have to operate on dense arrays per default.

If we convert everything to dense internally than we will not be able to handle MSMs with more than 4000 states in the MSM-object. If that is good for you then we have to redesign, but returning all time-scales will give a lot of stuff that doesn't make sense. 

I think we can prevent the average user from this mistake by explicitly asking for 'k' in the get_xxx functions. This will also allow us to use all the fast analysis code for sparse matrices in the api. For a given k<<n it we can return eigenvectors as dense arrays (I think this is even the default behavior of msm.analysis).

To warp up: I am still in favor to make k a mandatory keyword. In my opinion, all k eigenvectors rarely 'make sense'. A user who wants to do math with the full eigensystem should in my opnion do

T = msmob.tmatrix
T = T.toarray()
R = msm.analysis.eigenvectors(T)

because he only can judge if hewants to convert to dense.

gph82:
While we're at it, I think MSM.compute should already compute mu and some low number ( k = 10, sparse diagonalization) of EVs automatically. 

This way, the get_EVs function would be already reserved for users that, for some reason, want to go beyond k = 10 that at their own risk, and I suspect would be used less frequently

Now,  the code has to be like this:
```
MSM=pyemma.msm.ui.MSM(K10_disc_htheta, lagtimes[tau], compute=True)
MSM.get_LEVs(k=5)
MSM.get_REVs(k=5)
MSM.stationary_distribution
draw_kinmap(MSM.LEV[:,1], MSM.LEV[:,2], dot_size = MSM.mu)
```
while it is desirable something like
```
MSM=pyemma.msm.ui.MSM(K10_disc_htheta, lagtimes[tau], compute=True)
draw_kinmap(MSM.LEV[:,1], MSM.LEV[:,2], dot_size = MSM.mu)
```

Two related ideas:
- "draw_kinmap" could directly take the MSM object as argument (solves the above example, but the general case of wanting to plot the some EVs
- I know that Frank has plans to make the MSM object sort of a standard, MSM format, perhaps it's a good time to start converging to what I should be able to do

franknoe:
I agree.

I also suggest that plotting functions become more intelligent (and be 
included into emma!):

     draw_kinmap(MSM)

btw. in the code below it could just be

     MSM = pyemma.msm.msm(...)


Am 01/04/15 um 12:36 schrieb Guillermo Pérez-Hernández:
>
> While we're at it, I think MSM.compute should already compute mu and 
> some low number ( k = 10, sparse diagonalization) of EVs automatically.
>
> This way, the get_EVs function would be already reserved for users 
> that, for some reason, want to go beyond k = 10 that at their own 
> risk, and I suspect would be used less frequently
>
> Now, the code has to be like this:
>
> |MSM=pyemma.msm.ui.MSM(K10_disc_htheta, lagtimes[tau], compute=True)
> MSM.get_LEVs(k=5)
> MSM.get_REVs(k=5)
> MSM.stationary_distribution
> draw_kinmap(MSM.LEV[:,1], MSM.LEV[:,2], dot_size = MSM.mu)
> |
>
> while it is desirable something like
>
> |MSM=pyemma.msm.ui.MSM(K10_disc_htheta, lagtimes[tau], compute=True)
> draw_kinmap(MSM.LEV[:,1], MSM.LEV[:,2], dot_size = MSM.mu)
> |
>
> Two related ideas:
>
>   * "draw_kinmap" could directly take the MSM object as argument
>     (solves the above example, but the general case of wanting to plot
>     the some EVs
>   * I know that Frank has plans to make the MSM object sort of a
>     standard, MSM format, perhaps it's a good time to start converging
>     to what I should be able to do
>
> —
> <https://github.com/markovmodel/PyEMMA/pull/172#issuecomment-88435139>.
>


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

franknoe:
We could easily have a sparse=False option in the init. Then the number 
of eigenvalues computed must also be set. We can also generate a warning 
when too many states are asked for. And it should be possible to ask for 
eigenvalues / eigenvectors not to be computed by default.

At the high-level API I don't want any sparse matrices by default. 
Firstly because this is too complex for the average user. Secondly 
because solving eigenvalues with sparse matrices can often fail due to 
not convergence. With dense matrices I have never observed this problem.

Explicit computation of eigenvalues that have not yet be computed could 
be enforced by a function compute_eigen_decomposition(k = None). Getting 
the eigenvalues/eigenvectors and other properties should be handled by 
@property functions, rather than get_...() functions or direct access to 
the attributes.


Maybe I should say that the msm object is one that will go through major 
changes and extensions in the future to make msm estimation 
user-friendly, so we shouldn't be hesitant to make so little changes at 
this stage.

See also the msmbuilder MSM object for inspirations.



Am 01/04/15 um 11:52 schrieb Benjamin Trendelkamp-Schroer:
>
> Thanks Guillermo, you are correct. It makes no sense to use sparse 
> types and have methods that have to operate on dense arrays per default.
>
> If we convert everything to dense internally than we will not be able 
> to handle MSMs with more than 4000 states in the MSM-object. If that 
> is good for you then we have to redesign, but returning all 
> time-scales will give a lot of stuff that doesn't make sense.
>
> I think we can prevent the average user from this mistake by 
> explicitly asking for 'k' in the get_xxx functions. This will also 
> allow us to use all the fast analysis code for sparse matrices in the 
> api. For a given k<<n it we can return eigenvectors as dense arrays (I 
> think this is even the default behavior of msm.analysis).
>
> To warp up: I am still in favor to make k a mandatory keyword. In my 
> opinion, all k eigenvectors rarely 'make sense'. A user who wants to 
> do math with the full eigensystem should in my opnion do
>
> T = msmob.tmatrix
> T = T.toarray()
> R = msm.analysis.eigenvectors(T)
>
> because he only can judge if hewants to convert to dense.
>
> —
> <https://github.com/markovmodel/PyEMMA/pull/172#issuecomment-88419439>.
>


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
Ok, so in the spirit of "not being hesitant", I just did what so far we agree on on my last commit (3rd)

gph82:
The test fails with:
```build\temp.win32-2.7\Release\pyemma/coordinates/clustering/regspatial.obj /OUT:build\lib.win32-2.7\pyemma\coordinates\clustering\regspatial.pyd /IMPLIB:build\temp.win32-2.7\Release\pyemma/coordinates/clustering\regspatial.lib /MANIFESTFILE:build\temp.win32-2.7\Release\pyemma/coordinates/clustering\regspatial.pyd.manifest
LINK : fatal error LNK1181: cannot open input file 'theobald.lib'
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 9.0\\VC\\BIN\\link.exe' failed with exit status 1181
Command exited with code 1```

is this an mdtraj thing or ours?

marscher:
@gph82 Frank merged your branch locally and extended it and will send in a new PR soon.

gph82:
I thought it was desired that computing the MSM would automatically yield the first n-eigenvectors. The new MSM object does not do that. I still believe it should, regardless of the sparse/dense discussion at the api level.

franknoe:
It could be done, but do you feel that this is a limitation? Now the 
computation is done automatically as soon as you ask for them. Because 
the number of desired eigenvalues needs to be set for sparse mode, 
eigenvalues and eigenvectors have Arguments, so you need to call:
     msm.eigenvalues()
instead of just
     msm.eigenvalues

But otherwise the behavior is indistinguishable from you wanted, no?


Am 08/04/15 um 12:09 schrieb Guillermo Pérez-Hernández:
>
> I thought it was desired that computing the MSM would automatically 
> yield the first n-eigenvectors. The new MSM object does not do that. I 
> still believe it should, regardless of the sparse/dense discussion at 
> the api level.
>
> —
> <https://github.com/markovmodel/PyEMMA/pull/172#issuecomment-90869524>.
>


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
Kinda. Working on something now, will PR later. 

And yes, it's a bit limiting, at least in my view


--
Sent from my phone. 
Apologies for typos, unusual tone or brevity 

PR #177, opened 2015-04-01
Part of the coordinates.api.input(...) method, a few minor things

clonker:
- propagates max_clusters argument of the api into the regular space method
- small fix the the catch_unhashable method of the coordinates featurizer
- added way to instantiate the feature_reader by featurizer rather than just by input files and topology file
- input method of the coordinates api up to case 1.1

### DIFF
--- a/pyemma/coordinates/api.py
+++ b/pyemma/coordinates/api.py
@@ -11,17 +11,17 @@ from pyemma.util.log import getLogger
 
 from pyemma.coordinates.pipeline import Discretizer as _Discretizer
 # io
-from io.featurizer import MDFeaturizer as _MDFeaturizer
-from io.feature_reader import FeatureReader as _FeatureReader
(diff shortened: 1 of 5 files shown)


fabian-paul:
Hi Moritz,
could you please unify your implementation of load with coordinates.api.input, case 1? 
For input case 1, we would have exactly the same code safe for the final operate_in_memory, map_to_memory,...
I suggest to move your code into a new function and call this function in load and in input case 1.
Best, Fabian

fabian-paul:
Oops, I got this the wrong way round. Of course I mean: could you please extend your implementation of input such that it can be used in load?

marscher:
Thanks for the changes! Please move _get_file_reader to an impl package 
(maybe io.util) - We do not want impl detail in API module.

marscher:
superseeded by #180  (which includes this commits)

PR #181, opened 2015-04-02
[coor/transform] added linear operator. Fix #128

### LINKED ISSUE #128: Add option to give a linear operator to the transformation pipeline
This is a request in the context of hTICA, but might be otherwise also useful, perhaps.  

If I **already** have a transformation matrix and list of distances I want to operate on (s.t I rotate them into hTICs), this would be a useful feature, IMHO.

marscher:


franknoe:
Why do we need that?

franknoe:
+ Returns
+ -------
+ Y : ndarray(n,)
+ the projected data

cannot be true. It must be (T, n) where T is the number of frames in the chunk.

marscher:
yes the docstring is misleading. @gph82 has requested this feature, it may be useful to be able to perform arbitrary transformations given by a matrix. But one may argue, that it is too advanced to support it.

franknoe:
We should discuss with Guille what he wants to do with it. Moreover, 
this transformer can be generalized to an AffineTransform which would be 
more useful. Please close this PR for now and keep the branch.

Thanks, F.


-- 

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
I was not aware of Affine Transform at the moment of creating the request. For now, this is not critical.

PR #185, opened 2015-04-04
Imp save traj into refactor_discretizer branch

gph82:


gph82:
As to:

TODO: Is the result of coordinates.input the correct object here, or should we rather extract a separate loader class that is not featurized?

I don't think there is a problem with using an object that is featurized. We don't use the features, so what? As a matter of fact, we could even provide an option to save a "featurized" trajectory also (perhaps too convoluted? just a suggestion)

franknoe:
Tests are failing but this is not Guille's fault, because it's a featurizer problem. I will merge this for now as it's a feature branch.
@marscher have you already fixed these problem? Should I merge the lastest changes of devel into this branch?

PR #214, opened 2015-04-09
Kmeans quick fix and Pipeline.parametrize() stride fix

### LINKED ISSUE #196: [coordinates.api.pipeline]: Trying to pass argument to parametrize function
In line coordinates.api.pipeline the call:
        p.parametrize(param_stride)
is attempted, which results in an "Unexpected argument error" in PyCharm because Pipeline.parametrize does not accept arguments.

I'm not sure how to fix this, but if possible avoid passing arguments into parametrize.

franknoe:
* Quick Fix for k-means: Now accumulating data in memory and running full k-means from sklearn
* Fixes #196 
* New input checking for trajector data in coordinates.api. Uses new ensure_traj and ensure_trajs in util.types


franknoe:
Sorry, wrong target!

PR #217, opened 2015-04-11
Coor fixes 1.2

franknoe:
Fixes for the coordinates package. This PR looks big because of a package rename, but most changes are relatively minor. The most major change is using the coordinates.input (now coordinates.source) fixes and tests of @clonker.

Main changes:
* coordinates.input is fixed and can now construct FeatureReaders (raw data access not yet implemented)
* In coordinates.api.pca, tica, cluster_kmeans, etc: the construction of the transformer is simplified by calling a common helper function _param_stage
* In coordinates.api.pca, tica, cluster_kmeans, etc: we now also accept transformers as inputs. Thus you can do ad-hoc pipeline construction without using the api.pipeline function, e.g.:
```python
    import pyemma.coordinates as coor
    data_source = coor.source('lala.xtc', topology = 'blabla.pdb')
    tic = coor.tica(data_source)
```
   or alternatively in memory
```python
    import pyemma.coordinates as coor
    X= coor.load('lala.xtc', topology = 'blabla.pdb')
    tic = coor.tica(X)
```
* renamed coordinates.api.input to coordinates.api.source in order to avoid clash with python package
* renamed local attributes 'input' to 'inp' for the same reason
* renamed package coordinates.io to coordinates.data to avoid clash with python package (this affects many files)
* coordinates.clustering.uniform_time: the bug for large numbers of clusters is fixed (had nothing to do with the chunk size)


marscher:
looks good! Only my a test from my construction side is failing.

PR #224, opened 2015-04-11
[coordinates]: Removed some shadowing, minor bugs in tests, and fixed PE...

franknoe:
* Mostly cosmetic changes - PEP 8 coding style in coordinates package.
* Very few bugfixes (mostly trying to call TICA with nonexisting kwargs that were for some reason not causing failures so far)
* Removed some shadowing
* Nontrivial change: renamed kwargs topology -> top in order to be consistent with mdtraj
* Nontrivial change: renamed featurizer -> features in order to avoid shadowing


marscher:
I <3 pep8 :+1:

Issue #230, opened 2015-04-12
coordinates.cluster_assign_centers() doesn't work and isn't tested


franknoe:
```python
X = np.random.rand(100,2)
centers = np.array([[-0.5,0],[0.5,0]])
coor.cluster_assign_centers(data=X, centers=centers)
```

leads to:
```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-17-ad1f5ec290af> in <module>()
      1 X = np.random.rand(100,2)
      2 centers = np.array([[-0.5,0],[0.5,0]])
----> 3 coor.cluster_assign_centers(data=X, centers=centers)

/Users/noe/data/software_projects/pyemma/pyemma/coordinates/api.pyc in cluster_assign_centers(data, centers, stride)
    941                          ' or NumPy array')
    942     res = _AssignCenters(centers)
--> 943     return _param_stage(data, res, stride=stride)

/Users/noe/data/software_projects/pyemma/pyemma/coordinates/api.pyc in _param_stage(previous_stage, this_stage, stride)
    585     this_stage.data_producer = inputstage
    586     this_stage.chunksize = inputstage.chunksize
--> 587     this_stage.parametrize(stride=stride)
    588     return this_stage
    589 

/Users/noe/data/software_projects/pyemma/pyemma/coordinates/transform/transformer.pyc in parametrize(self, stride)
    234                     # first chunk
    235                     return_value = self._param_add_data(
--> 236                         X, itraj, t, first_chunk, last_chunk_in_traj, last_chunk, ipass, Y=Y, stride=stride)
    237                     if isinstance(return_value, tuple):
    238                         add_data_finished, lag = return_value

/Users/noe/data/software_projects/pyemma/pyemma/coordinates/clustering/assign.pyc in _param_add_data(self, X, itraj, t, first_chunk, last_chunk_in_traj, last_chunk, ipass, Y, stride)
     45         if t == 0:
     46             n = self.data_producer.trajectory_length(itraj, stride=stride)
---> 47             self.dtrajs.append(np.empty(n, dtype=int))
     48 
     49         L = np.shape(X)[0]

/Users/noe/data/software_projects/pyemma/pyemma/coordinates/clustering/interface.pyc in dtrajs(self)
     29     def dtrajs(self):
     30         if (len(self._dtrajs) == 0):  # nothing assigned yet, doing that now
---> 31             return self.assign(stride = 1)
     32         else:
[...]

PR #236, opened 2015-04-13
Clustering fixes

marscher:
* fix for #230 
* cleanup regspace test unittest


franknoe:
That's the right way to fix it. Go ahead and merge.

PR #241, opened 2015-04-13
Fix windows problems

fabian-paul:
don't merge yet


pyemma/msm/analysis/dense/decomposition.py
```diff
@@ -189,7 +190,7 @@ def eigenvectors(T, k=None, right=True):
     """ Return eigenvectors """
     if k is None:
         return eigvec
-    elif isinstance(k, int):
+    elif isinstance(k, numbers.Integral):
```
marscher: consider using util.types.is_int here

PR #244, opened 2015-04-13
[coor/api/assign_to_centers] by default return dtrajs. Added flag to opt...

marscher:
added optional flag to api function assign_to_centers to return assignment instance.


franknoe:
Looks good, go ahead.

PR #278, opened 2015-04-29
transposes the returned eigenvectors if right=False

### LINKED ISSUE #249: left eigenvectors...rows or columns?
Hi, I believe we discussed this issue yesterday. 

I brought it up because I remembered that left eigenvectors could be returned either as rows or columns depending on how they were computed. Shouldn't we unify this?

```
T is of class  <class 'pyemma.msm.ui.msm.MSM'>
T.transition_matrix
[[ 0.999  0.001  0.     0.   ]
 [ 0.002  0.995  0.003  0.   ]
 [ 0.     0.001  0.997  0.002]
 [ 0.001  0.001  0.5    0.498]]

T.eigenvectors_left()
[[  3.35309618e-01   1.66666667e-01   4.96047431e-01   1.97628458e-03]
 [ -8.24131911e-01  -1.82882024e-13   8.20855019e-01   3.27689235e-03]
 [  3.52341480e-01  -8.81917104e-01   5.27448814e-01   2.12680974e-03]
 [  1.98021364e-03  -2.04743344e-15   9.94059375e-01  -9.96039589e-01]]

pyemma.msm.analysis.eigenvectors(T.transition_matrix, right = False)
[[  5.39506188e-01  -7.08511154e-01   3.24338885e-01   1.40718672e-03]
 [  2.68163194e-01  -1.57148555e-13  -8.11826102e-01  -8.53499247e-16]
 [  7.98129979e-01   7.05693990e-01   4.85529437e-01   7.06402138e-01]
 [  3.17980071e-03   2.81716404e-03   1.95777999e-03  -7.07809324e-01]]
```

I've specifically avoided  any normalization so that we also realise that they're being returned with different norms. The graph shows that the first row of the first matrix is actually the first column of the second matrix times a constant
rows-vs-columns

gph82:
This should close #249 The tests have been updated accordingly. Pycharm reports no other usages except for the declaration in the api, so nothing else should be broken (please @trendelkampschroer , can you confirm?) 

Do we include this in the changelog?

### DIFF
--- a/pyemma/msm/analysis/api.py
+++ b/pyemma/msm/analysis/api.py
@@ -503,6 +503,9 @@ def eigenvectors(T, k=None, right=True, ncv=None):
         the corresponding eigenvalue. If k is None then n=d, if k is
         int then n=k.
 
+        If right = True, the right eigenvectors are returned as column vectors.
+        If right = False, the left eigenvectors are returned as row vectors
+
     See also
     --------
     rdl_decomposition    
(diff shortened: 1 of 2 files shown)


trendelkampschroer:
Confirmed.

PR #277, opened 2015-04-29
Include tica correlations

gph82:
closes #268

PR #280, opened 2015-04-29
Kmeans c extension

clonker:
Adds a kmeans c extension and changes some minor things in the clustering interface.


marscher:
can you please rebase against devel and change base against dev_release_1.2.1 (as discussed in #277)

franknoe:
I'm not sure I understand the code. Why is a kmeans_clustering.cluster called every time add_data is called? Shouldn't that happen only once?

Please try this with real data (with respect to behavior and performance). You can use the London workshop notebook solution.ipynb found with all data under:
https://www.dropbox.com/sh/nftmtyybur5x0i5/AADaD7pomY_BfWldCGfzviGFa?dl=0

PR #287, opened 2015-04-30
Update api.py

ismaelresp:
'fmt' option in pyemma.coordinates.save_trajs(), missing dot in output filenames when 'fmt' is not by default.


marscher:
Thank you!

Issue #183, opened 2015-04-03
Unify error/warning behavior in reversible transition matrix estimators


franknoe:
Fabian's and my implementation has currently different behavior with respect to convergence failure. In Fab's impl, an error is raised when no convergence is achieved, or even when there is the danger of not achieving convergence. In my impl. there is no such error, but it's possible to check a posteriori.

I would suggest to generate a warning when not converged, but still to return a result.

PR #289, opened 2015-04-30
Started to unify error conditions in reversible matrix estimation

fabian-paul:
solves issue #183
When no convergence is reached, a NotConvergedWarning is emited via the Python warning mechanism.


franknoe:
OK.

Martin: please don't merge so fast. I want to have a chance to look at PR's. We can wait for one day because there is no urgent deadline close.

PR #290, opened 2015-04-30
[coor/feature_reader|config] cache trajectory lengths in (optional) cache.

### LINKED ISSUE #279: FeatureReader should (optionally) create an index cache (trajectory lengths)
This would avoid running over all trajs initially.
For dimensions we would only need to read the first frame. The index could be created as a hidden file (eg. .traj_x.index.txt) and it could be automatically checked if it exists. We should add a timestamp, when this lengths information has been generated to avoid the case of concatenated trajs being reported as the old cached length.

marscher:
Can be turned of via config. A default config is now created in ~/.pyemma/

Fixes #279


franknoe:
Can we discuss it on monday before merging? I don't understand yet what this does.

PR #291, opened 2015-04-30
[coor/data/featurizer] made it a Transformer

marscher:


franknoe:
Why make it a transformer? I understand that in principle it would be one, but the way we have implemented it, it requires an mdtraj Trajectory as an input and produces an ndarray, while transformers always map arrays to arrays. Also there is nothing else in the transformer class that is used here it seems.

PR #304, opened 2015-05-12
Update chapman_kolmogorov.py (1)

ismaelresp:
Added reversible option for estimated MD transition matrix.


franknoe:
I am afraid this is not meaningful. The cktest should compare a model (transition matrix, estimated in any arbitrary way) to the direct trajectory observation. We just use the nonreversible transition matrix estimator to compute the directly observed fraction of transitions. Replacing this with a reversible estimator would mean that you compare two models with one another, which is not the point of the cktest.

PR #316, opened 2015-05-20
Improvements featurizer

sstolzenberg:
- pre-selection for backbone torsions.
- cos/sin components for chi1/phi/psi.
- labels

### DIFF
--- a/pyemma/coordinates/data/featurizer.py
+++ b/pyemma/coordinates/data/featurizer.py
@@ -25,7 +25,7 @@
 __author__ = 'Frank Noe, Martin Scherer'
 
 import mdtraj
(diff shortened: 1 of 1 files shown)


franknoe:
- If we do it, cos/sin calculation should be added to all torsion features, not just backbone and chi1torsions. So it should also appear in DihedralFeature and in add_dihedrals. Maybe we can put that functionality in the DihedralFeature superclass and just manage the dihedral selection in the subclasses?
- The cos/sin option should affect the describe() output as well, because you now have doubled your dimensions, so you also need to double the number of description strings. For example, one could use "COS ..." and "SIN ..." followed by the normal dihedral description. Maybe you are doing something like this, but I can't spot it.

marscher:
merged this one into #370

PR #327, opened 2015-05-23
Fast timescale calculation

franknoe:
The timescale calculation is sped up a lot due to two changes:
- Now converting to a dense matrix just before calculating the timescales. This way we do the reversible estimation on a sparse matrix which is for some reason much faster, although I don't understand why because the sparsity pattern is the same and should be exploited also in the dense implementation ( @fabian-paul , you said that for the dense T-matrix estimation a C-extension would not speed it up, but this seems to contradict my observation. )
- Added fast eigenvalue and timescale calculation for reversible T-matrices to msm.analysis. There is an option reversible=False and an option mu=None for the stationary distribution. If reversible=True, then the stationary distribution is used or computed to symmetrise the T-matrix as s_ij = sqrt(mu_i/mu_j) T_ij. Then eigenvalsh is used to compute the eigenvalues (identical problem, but much faster).

I would like @trendelkampschroer to check this PR. @trendelkampschroer , if you agree with these changes please do a corresponding enhancement for the eigenvectors (which are the eigenvectors of S, scaled by 1/sqrt(mu)) and rdl decomposition and also for the sparse case.

### DIFF
--- a/pyemma/msm/analysis/api.py
+++ b/pyemma/msm/analysis/api.py
@@ -390,7 +390,7 @@ def stationary_distribution(T):
 
 
 # DONE: Martin, Ben
-def eigenvalues(T, k=None, ncv=None):
+def eigenvalues(T, k=None, ncv=None, reversible=False, mu=None):
     r"""Find eigenvalues of the transition matrix.
 
     Parameters
@@ -402,6 +402,12 @@ def eigenvalues(T, k=None, ncv=None):
     ncv : int (optional)
         The number of Lanczos vectors generated, `ncv` must be greater than k;
         it is recommended that ncv > 2*k
+    reversible : bool (optional)
+        Indicate that transition matrix is reversible. Will compute its stationary distribution `\mu` (unless given)
+        and then compute the eigenvalues of the symmetric matrix `\sqrt(\mu_i / \mu_j)` which is equivalent but
+        much faster
+    mu : numpy.ndarray, shape=(d)
+        Stationary distribution of T. Will only be used if reversible=True in order to symmetrize T.
 
     Returns
     -------
@@ -427,13 +433,13 @@ def eigenvalues(T, k=None, ncv=None):
     if _issparse(T):
         return sparse.decomposition.eigenvalues(T, k, ncv=ncv)
     elif _isdense(T):
-        return dense.decomposition.eigenvalues(T, k)
+        return dense.decomposition.eigenvalues(T, k, reversible=reversible, mu=mu)
     else:
         raise _type_not_supported
 
 
 # DONE: Ben
-def timescales(T, tau=1, k=None, ncv=None):
+def timescales(T, tau=1, k=None, ncv=None, reversible=False, mu=None):
     r"""Compute implied time scales of given transition matrix.
 
     Parameters
@@ -449,6 +455,12 @@ def timescales(T, tau=1, k=None, ncv=None):
     ncv : int (optional, for sparse T only)
         The number of Lanczos vectors generated, `ncv` must be greater than k;
         it is recommended that ncv > 2*k
+    reversible : bool (optional)
+        Indicate that transition matrix is reversible. Will compute its stationary distribution `\mu` (unless given)
+        and then compute the eigenvalues of the symmetric matrix `\sqrt(\mu_i / \mu_j)` which is equivalent but
+        much faster
+    mu : numpy.ndarray, shape=(d)
+        Stationary distribution of T. Will only be used if reversible=True in order to symmetrize T.
 
     Returns
     -------
@@ -476,7 +488,7 @@ def timescales(T, tau=1, k=None, ncv=None):
     if _issparse(T):
         return sparse.decomposition.timescales(T, tau=tau, k=k, ncv=ncv)
     elif _isdense(T):
-        return dense.decomposition.timescales(T, tau=tau, k=k)
+        return dense.decomposition.timescales(T, tau=tau, k=k, reversible=reversible, mu=mu)
     else:
         raise _type_not_supported
 
--- a/pyemma/msm/analysis/dense/decomposition.py
+++ b/pyemma/msm/analysis/dense/decomposition.py
@@ -37,7 +37,7 @@ import numpy as np
 import numbers
 import warnings
 
-from scipy.linalg import eig, eigvals, solve, lu_factor, lu_solve
+from scipy.linalg import eig, eigh, eigvals, eigvalsh, solve, lu_factor, lu_solve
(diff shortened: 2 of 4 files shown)


franknoe:
I don't understand the failing test. TICA does not use any of the functions I have changed, because they are exclusively relevant for transition matrices. Maybe this problem existed before or is non-deterministic?

fabian-paul:
> This way we do the reversible estimation on a sparse matrix which is for some reason much faster, although I don't understand why because the sparsity pattern is the same and should be exploited also in the dense implementation

I don't understand, could you explain this? How is the sparsity pattern pattern exploited in the dense implementation?

> ( @fabian-paul , you said that for the dense T-matrix estimation a C-extension would not speed it up, but this seems to contradict my observation. )

This was my observation that I made during reversible estimations of T matrices with fixed stationary distribution. I implemented both sparse and dense versions as C extensions in addition to your dense Python-implementation that was already in place. The dense C implementation did not show any substantial speed-up compared to the dense Python implementation.
If I remember correctly, the dense Python implementation is quite optimized, it recycles numpy arrays etc. Maybe we should do the same with the dense numpy implementation without pi.

franknoe:
Am 26/05/15 um 10:55 schrieb fabian-paul:

> ```
> This way we do the reversible estimation on a sparse matrix which
> is for some reason much faster, although I don't understand why
> because the sparsity pattern is the same and should be exploited
> also in the dense implementation
> ```
> 
> I don't understand, could you explain this? How is the sparsity 
> pattern pattern exploited in the dense implementation?
> 
> It uses np.nonzero of C+C.T, because we know that the solution is zero 
> everywhere else. Effectively it works on a sparse matrix but perhaps the 
> implementation is less efficient.
> 
> ```
> ( @fabian-paul <https://github.com/fabian-paul> , you said that
> for the dense T-matrix estimation a C-extension would not speed it
> up, but this seems to contradict my observation. )
> ```
> 
> This was my observation that I made during reversible estimations of T 
> matrices with fixed stationary distribution. I implemented both sparse 
> and dense versions as C extensions in addition to your dense 
> Python-implementation that was already in place. The dense C 
> implementation did not show any substantial speed-up compared to the 
> dense Python implementation.
> 
> Perhaps the sparsity pattern was not exploited?
> 
> If I remember correctly, the dense Python implementation is quite 
> optimized, it recycles numpy arrays etc. Maybe we should do the same 
> with the dense numpy implementation without pi.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/327#issuecomment-105452675.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

trendelkampschroer:
@franknoe: This is a very nice improvement. What speedups did you observe using eigvalsh instead of eigvals?

 Can you explain why the matrix S = D_{smu} T D_{smu}^{-1} is always symmetric for a T that fulfills detailed balance. I have never been able to show this formally and would like to know how that works.

I would try to circumvent a division by smu. Since mu and also smu can contain small numbers and I belive that small entries on mu can carry errors (statistical as well as numerical) we should think about another way.

I can prepare a fix, using a generalized eigenvalue problem instead, that way we do not need to divide by smu.

franknoe:
Am 27/05/15 um 12:11 schrieb Benjamin Trendelkamp-Schroer:

> @franknoe https://github.com/franknoe: This is a very nice 
> improvement. What speedups did you observe using eigvalsh instead of 
> eigvals?
> 
> For the pure eigenvalue calculation and large matrices it can be 
> massive, e.g. a factor 100 for a 5000x5000 matrix

For the implied timescales calculation often the reversible transition 
matrix estimation is more expensive, so it's difficult to answer there. 
This is why I also changed the way that the estimation is done in 
implied timescales to save at that end.

> Can you explain why the matrix S = D_{smu} T D_{smu}^{-1} is always 
> symmetric for a T that fulfills detailed balance. I have never been 
> able to show this formally and would like to know how that works.
> 
> Insert the definition of S
>      s_ij = sqrt(pi_i/pi_j) p_ij.
> into detailed balance
>      pi_i p_ij = pi_j pi_ji
> and you'll get
>      s_ij = s_ji.
> 
> I would try to circumvent a division by smu. Since mu and also smu can 
> contain small numbers and I belive that small entries on mu can carry 
> errors (statistical as well as numerical) we should think about 
> another way.
> 
> I'm not sure if this is a problem. We know mu is not zero at the point 
> we're using this. Then the smallest possible mu is 1e-16 and the 
> smallest possible smu is 1e-8, and the largest possible sqrt(pi_i/pi_j) 
> would be 1e8. This does not seem to be a problem.
> 
> I can prepare a fix, using a generalized eigenvalue problem instead, 
> that way we do not need to divide by smu.
> 
> You'd have to demonstrate that this is still as fast as with eigvalsh, 
> because the whole point of this PR is a speedup. Note that eigvalsh is 
> much faster than eigh for example.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/327#issuecomment-105852998.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
Is this in a mergeable state yet? If yes I really want to include it in the 1.2.1 release.

franknoe:
I think yes, but let's wait for Ben's comment.

Am 27/05/15 um 17:09 schrieb Martin K. Scherer:

> Is this in a mergeable state yet? If yes I really want to include it 
> in the 1.2.1 release.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/327#issuecomment-105949911.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

trendelkampschroer:
@franknoe : If you are confident about the division by the stationary distribution and all tests are successful than I think this can be merged.

Thanks for this nice improvement!

franknoe:
Yes, all tests are working. Thank you!

Martin: go ahead. Maybe we should now make a freeze for the 1.2.1 
release and compile the changelog.

Am 27/05/15 um 17:29 schrieb Benjamin Trendelkamp-Schroer:

> @franknoe https://github.com/franknoe : If you are confident about 
> the division by the stationary distribution and all tests are 
> successful than I think this can be merged.
> 
> Thanks for this nice improvement!
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/327#issuecomment-105959712.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
Alright. I've already started the changelog for 1.2.1, but lets check what is missing yet.

Issue #340, opened 2015-06-02
Transformer.trajectory_length(s) should return numpy arrays


marscher:
That allows to do algebra on these quantities more easily. It will not brake the api, since a numpy array and a list are essentially api compatible from a duck typing point of view. This has been suggested by @gph82 recently.

franknoe:
OK

Am 02/06/15 um 06:17 schrieb Martin K. Scherer:

> That allows to do algebra on these quantities more easily. It will not 
> brake the api, since a numpy array and a list are essentially api 
> compatible from a duck typing point of view. This has been suggested 
> by @gph82 https://github.com/gph82 recently.
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/markovmodel/PyEMMA/issues/340.

## 

---

Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

## Mail:  Arnimallee 6, 14195 Berlin, Germany

gph82:
+1

PR #341, opened 2015-06-02
[msm.api.its] gets a new property "fraction_of_frames" (for every lagtime)

gph82:
I think it is useful to have this when dealing with dtrajs of significantly different lenghts.

TODO: I would also want to add sth like fraction_active_frames. 

I updated the docs, and also the ./doc/source/api/generated/pyemma.msm.ui.ImpliedTimescales.rst file, but it is not tracked. Do we want to add them?

PR #342, opened 2015-06-03
[transformer] method trajectory_lengths returns numpy int array now. 

marscher:
This implements feature requested in #340


franknoe:
Looks good but appveyor is failing.

gph82:
Appears again to be an ImaginaryEigenValueWarning before the assertion error. Can this be related to why #341 is also failing?

marscher:
Yes thats the random failing TICA test again. The chosen seed only works under linux... We should maybe think of a proper way how to fix that test.

PR #343, opened 2015-06-03
added a minRMSD feature.

gph82:
./source/api/generated/pyemma.coordinates.data.MDFeaturizer.rst has also been changed, but we're not tracking these files

There is a TODO regarding duplicates when parsing the same trajectory through different rajectory objects. The hashed trajectory objects are different.

Also, pretty verbose feat.describe() output, it is a tentative format.

### DIFF
--- a/pyemma/coordinates/data/featurizer.py
+++ b/pyemma/coordinates/data/featurizer.py
@@ -386,6 +386,65 @@ class BackboneTorsionFeature(object):
     def __eq__(self, other):
         return self.__hash__() == other.__hash__()
 
+class MinRmsdFeature(object):
+
+    def __init__(self, ref, ref_frame=0, atom_indices=None, topology=None, precentered=False):
+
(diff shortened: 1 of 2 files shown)


franknoe:
Looks good. Thanks.

marscher:
Looks very good. The hashing stuff is not that an important issue. Trajectory object do not provide a custom hashing function, thats why the most basic implementation of object is being called, which causes collisions. One might consider sending them a PR, to implement proper hashing (it might be interesting if you want to lookup e.g Trajectory objects in a cache really fast).

If you want to fix this behaviour, you can write your own hash function for Trajectory (which should consider the topology, coords and time arrays) and call it in the features hash function.

marscher:
send in a PR to mdtraj

gph82:
Thanks for the quick merge. 

@marscher , what do we do with ./source/api/generated/pyemma.coordinates.data.MDFeaturizer.rst and the like?

marscher:
Sorry, I do not understand your question at all.

gph82:
You can ignore it safely ;)

PR #345, opened 2015-06-05
Fix chunksize datainmemory

marscher:


fabian-paul:
The second example in the issue is still not working.

PR #347, opened 2015-06-05
Fix chunksize datainmemory

fabian-paul:
Your fix isn't working. I changed it.
Now the following works:

``` Python
import pyemma
import numpy as np
data1 = np.random.randn(12345,10)
data2 = np.random.randn(54321,10)
source = pyemma.coordinates.source([data1,data2])
stride = 123
lag = 45
source.chunksize = 678
c = 0
c_lag = 0
for i,X,Y in source.iterator(stride=stride,lag=lag):
    c+=X.shape[0]
    c_lag+=Y.shape[0]
print 'got', c, 'expected', ((data1.shape[0]-1)//stride+1) + ((data2.shape[0]-1)//stride+1)
print 'got', c_lag, 'expected', ((data1.shape[0]-1)//stride+1 - lag) +  ((data2.shape[0]-1)//stride+1 - lag) # old convention

```

### DIFF
--- a/pyemma/coordinates/data/data_in_memory.py
+++ b/pyemma/coordinates/data/data_in_memory.py
@@ -133,22 +133,19 @@ class DataInMemory(ReaderInterface):
 
             X = traj[slice_x]
 
-            if lag == 0:
-                self._t = upper_bound
+            if lag!=0:
+                 upper_bound_Y = min(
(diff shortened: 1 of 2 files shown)


marscher:
Thank you! Unfortunately I accidentally committed changes to pyemma-ipython submodule, which are not yet upstream. I think I'm gonna push them now and assume you have tested this, so we are ready to merge.

Issue #353, opened 2015-06-11
simplify pyemma.coordinates.tests.test_stride.TestStride 


franknoe:
This test takes very long and clutters the test output. I'm sure it can be simplified.

fabian-paul:
I would like to fix this but I don't know how. 
Is it possible to temporarily suppress displaying of the progress bar and log messages? 
I would like to write something like:

``` Python
with no_logging:
   t = tica(...)
   ...
```

Speeding up the test is easily done by reducing the range of strides that the function tests and by reducing the dimensions of the test data. 
Optimally, I would like to fix both issues in one PR.

franknoe:
Am 12/06/15 um 10:33 schrieb fabian-paul:

> I would like to fix this but I don't know how.
> 
> I understood the problem. Sorry, this is from me, I was under the 
> impression it might have been one of your tests.
> The reason is that the test computes the wrong reference value by not 
> symmetrizing C(tau).
> 
> Is it possible to temporarily suppress displaying of the progress bar 
> and log messages?
> 
> I'm sure it is, but Martin or Moritz need to tell us how.
> 
> I would like to write something like:
> 
> with  no_logging:
>     t=  tica(...)
>     ...
> 
> Speeding up the test is easily done by reducing the range of strides 
> that the function tests and by reducing the dimensions of the test data.
> Optimally, I would like to fix both issues in one PR.
> 
> OK
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/markovmodel/PyEMMA/issues/353#issuecomment-111413343.

## 

---

Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

## Mail:  Arnimallee 6, 14195 Berlin, Germany

clonker:
It is a config value

```
show_progress_bars = True
```

so setting this to False should disable the progress bars. If it is desired to disable them runtime-temporarily, try something like this:

``` python
from pyemma.util.config import conf_values

# to disable
conf_values['pyemma'].show_progress_bars = 'False'
# to enable
conf_values['pyemma'].show_progress_bars = 'True'
```

franknoe:
I'm a bit confused with these config values. As far as I understand the 
config values are quasi-permanent, i.e. they are set for a certain user 
(in a config file), and then every time I set their value, I change the 
config file. Right?

[...]

Issue #354, opened 2015-06-11
test_networks.TestNetworkPlot plots


franknoe:
The test

test_networks.TestNetworkPlot

actually plots a figure during the test run. I'm sure that's not desired.

marscher:
There seems no good possibility to switch the backend of matplotlib after it has been imported. The backend should be headless, so no GUI will popup during testing. But I would strongly recommend to test plots at least for exception safety (NetworkPlot eg had tons of failures due to wrong imports of grandalf library).

franknoe:
We could do the following: use an option draw=True which triggers the 
plt.show() in the plotting function. In the test, use the plotting 
function with draw=False.

Grandalf-wise: I would be happy to switch to networkx if that's a better 
option. But I think this is more a cosmetic and could be moved until 
after the release that we use for the paper.

Am 15/06/15 um 12:51 schrieb Martin K. Scherer:

> There seems no good possibility to switch the backend of matplotlib 
> after it has been imported. The backend should be headless, so no GUI 
> will popup during testing. But I would strongly recommend to test 
> plots at least for exception safety (NetworkPlot eg had tons of 
> failures due to wrong imports of grandalf library).
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/markovmodel/PyEMMA/issues/354#issuecomment-112016914.

## 

---

Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

## Mail:  Arnimallee 6, 14195 Berlin, Germany

marscher:
Am 15.06.2015 um 14:48 schrieb Frank Noe:

> We could do the following: use an option draw=True which triggers the
> plt.show() in the plotting function. In the test, use the plotting
> function with draw=False.

Thats a good idea! In principle do we always return an Axes object or do
we always call show in our plot functions? I'm not sure if that is
consistently done in all of them.

If we always return the Axes or Figure object it is up to user what he
will do with it (and he also be able to modify it).

franknoe:
Am 15/06/15 um 14:53 schrieb Martin K. Scherer:

> Am 15.06.2015 um 14:48 schrieb Frank Noe:
> 
> > We could do the following: use an option draw=True which triggers the
> > plt.show() in the plotting function. In the test, use the plotting
> > function with draw=False.
> 
> Thats a good idea! In principle do we always return an Axes object or do
[...]

PR #357, opened 2015-06-15
[coordinates][tests] speeded-up test_stride and suppressed displaying of progress bars

fabian-paul:
This is my suggestion for fixing #353.
The dimensions of the test data was reduced significantly. This results in a  run time of 3 seconds (before it was 12 seconds).
Now progress bars are not displayed during the test, this keeps the console output smaller. 
Logger messages are still displayed. I don't know how to switch them off.

### DIFF
--- a/pyemma/coordinates/tests/test_stride.py
+++ b/pyemma/coordinates/tests/test_stride.py
@@ -30,17 +30,6 @@ import numpy as np
 import mdtraj
 import pyemma.coordinates as coor
 
-from pyemma.util.config import conf_values
-class NoProgressBars:
-    def __enter__(self):
-        self._show_progress_bars = conf_values['pyemma'].show_progress_bars
-        conf_values['pyemma'].show_progress_bars = 'False'
-        return self
-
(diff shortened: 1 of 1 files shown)


marscher:
Thanks for the new and faster test. I think problems like displaying 
progressbars and having plots displayed during testing arises by not 
running the tests via nosetests, which now use quite a while to setup a 
common environment to run them. (see #354, of course it is not desired.)

So we should require to use nose in the documentation. At least it is 
required for testing subpackages coordinates and plots. I would also be 
happy not to enforce it, but then we need a better option to switch the 
plotting backend in matplotlib and hiding progressbars (nobody wants to 
use a context manager within every testcase!).

I like the context manager for switching the config values though!

fabian-paul:
I agree, using a context manager in every test is awkward. I'm open to better solutions.

marscher:
the option is to use nosetests so far. If nobody has a better one, we 
should stick to that.

fabian-paul:
Ok, I've taken out the context managers again.

PR #360, opened 2015-06-15
[util/progressbar] check if session is attached to a tty 

marscher:
or has a main function before displaying progress.

The default remains to show them by default, since they can be switched off during runtime.


franknoe:
ok

PR #361, opened 2015-06-15
Fix mle trev given pi

fabian-paul:
I made come changes to the reversible estimators with fixed pi:
- all estimators now accept non-integer counts
- the regularization parameter is removed from the implementations because it is not necessary (it was actually only there, because we made an error in deriving the estimator); If somebody wants the parameter back for some reason, this can be achieved by simply adding the parameter to the count matrix explicitly (this works now, because the count matrix is of type double.)


franknoe:
Looks good!

fabian-paul:
Anybody volunteers to review the PR?

PR #363, opened 2015-06-15
Feature auto sparse

fabian-paul:
This implements a feature I discussed with Frank last week: automatic choice of sparse or dense algorithm for ML T matrix estimation. Whichever algorithm is more efficient, is selected.
The criterion is a simple heuristic that compares the potential memory usage of the sparse and the dense representations of the C (or T) matrix. If 3*n_dof < dim the sparse algorithm is used. (n_dof is the number non-zero elements in the C matrix and dim is its dimension). 3*n_dof > dim, dense is used.
The idea behind the heuristic is that in many cases, the speed of computation is limited by the access to the data. Therefore the more compact data representation should yield a faster algorithm.
The new behavior can be controlled by the auto_sparse parameter of pyemma.msm.estimation.tmatrix and is on by default.

### DIFF
--- a/pyemma/msm/estimation/api.py
+++ b/pyemma/msm/estimation/api.py
@@ -38,6 +38,7 @@ import warnings
 import numpy as np
 
 from scipy.sparse import csr_matrix
+from scipy.sparse import coo_matrix
 from scipy.sparse import issparse
 from scipy.sparse.sputils import isdense
 
@@ -725,7 +726,7 @@ def prior_rev(C, alpha=-1.0):
 ################################################################################
 
 @shortcut('tmatrix')
-def transition_matrix(C, reversible=False, mu=None, **kwargs):
+def transition_matrix(C, reversible=False, mu=None, auto_sparse=True, **kwargs):
     r"""Estimate the transition matrix from the given countmatrix.   
     
     Parameters
@@ -739,9 +740,15 @@ def transition_matrix(C, reversible=False, mu=None, **kwargs):
         space of stochastic matrices.
     mu : array_like
         The stationary distribution of the MLE transition matrix.
-    **kwargs:
-        Optional algorithm-specific parameters. See below for special cases
-    eps : 1E-6 : float
+    auto_sparse : bool (optional, default=True)
+        Tf true: automatically choose the sparse or the dense
+        implementation, whichever is more efficient. If false, choose
+        the implementation that matches the input type.
+        The criterion to select the sparse implementation is 3*n<dim^2
+        where n is the number of non-zero elements in C and dim is the
+        dimension of C.
+    **kwargs: Optional algorithm-specific parameters. See below for special cases
+    eps = 1E-6 : float
         Optional parameter with reversible = True and mu!=None.
         Regularization parameter for the interior point method. This value is added
         to the diagonal elements of C that are zero.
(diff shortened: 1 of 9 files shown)


franknoe:
Thanks Fab. I'll review this.

Am 15/06/15 um 19:17 schrieb fabian-paul:

> This implements a feature I discussed with Frank last week: automatic 
> choice of sparse or dense algorithm for ML T matrix estimation. 
> Whichever algorithm is more efficient, is selected.
> The criterion is a simple heuristic that compares the potential memory 
> usage of the sparse and the dense representations of the C (or T) 
> matrix. If 3_n_dof < dim the sparse algorithm is used. (n_dof is the 
> number non-zero elements in the C matrix and dim is its dimension). 
> 3_n_dof > dim, dense is used.
> The idea behind the heuristic is that in many cases, the speed of 
> computation is limited by the access to the data. Therefore the more 
> compact data representation should yield a faster algorithm.
> The new behavior can be controlled by the auto_sparse parameter of 
> pyemma.msm.estimation.tmatrix and is on by default.
> 
> ---
> 
> ```
>     You can view, comment on, or merge this pull request online at:
> ```
> 
> https://github.com/markovmodel/PyEMMA/pull/363
> 
> ```
>     Commit Summary
> ```
> - [msm][estimation] removed regularization constant from sparse
>   reversible ML T estimator with fixed pi (because it isn't needed)
> - [msm][estimation] added deprecation warning and adapted test
> - [msm][estimation] dense reversible ML T matrix estiamtor with
>   fixed pi now accepts non-integer counts; regularization parameter
>   removed
> - [msm][estimation] new feature: automatically choose the more
>   efficient of sparse or dense in ML T matrix estimation.
>   
>   ```
>   File Changes
>   ```
> - _M_ pyemma/msm/estimation/api.py
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-0 (62)
> - _M_ pyemma/msm/estimation/dense/_mle_trev_given_pi.c
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-1 (55)
> - _M_ pyemma/msm/estimation/dense/_mle_trev_given_pi.h
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-2 (3)
> - _M_ pyemma/msm/estimation/dense/mle_trev_given_pi.pyx
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-3 (25)
> - _M_ pyemma/msm/estimation/sparse/_mle_trev_given_pi.c
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-4 (23)
> - _M_ pyemma/msm/estimation/sparse/_mle_trev_given_pi.h
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-5 (2)
> - _M_ pyemma/msm/estimation/sparse/mle_trev_given_pi.pyx
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-6 (87)
> - _M_ pyemma/msm/estimation/tests/test_mle_trev_given_pi.py
>   https://github.com/markovmodel/PyEMMA/pull/363/files#diff-7 (15)
>   
>   ```
>   Patch Links:
>   ```
> - https://github.com/markovmodel/PyEMMA/pull/363.patch
> - https://github.com/markovmodel/PyEMMA/pull/363.diff
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/363.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

franknoe:
Thanks! Very good work. I just have a few cosmetic suggestions:
- I find the parameter auto_sparse=True unclear. It is really meant to control the computation method and has nothing to do with the data type. For selection different computation methods, I generally suggest to use the parameter name "method", which could here take values 'auto' or 'dense' or 'sparse'. So the default would be method='auto'
- Your description of when you switch is unclear. First it's dimension**2 rather than dimension that is used. Then the bottomline is basicly that you use sparse calculation if less than 1/3 of the elements are nonzero, and dense otherwise. So I would write it this way.

If you just push again with these changes I or martin can merge.

fabian-paul:
>    I find the parameter auto_sparse=True unclear. It is really meant to control the computation method and has nothing to do with the data type. For selection different computation methods, I generally suggest to use the parameter name "method", which could here take values 'auto' or 'dense' or 'sparse'. So the default would be method='auto'

I think there should also be a value of method='input_type' that reproduces the current behavior of the method. Right now (before my changes), tmatrix() uses the sparse algorithm if the input type is sparse and the dense algorithm if the input type is dense. It might be useful to provide this behavior for the user who needs it.
Then I also would not name the method of choosing the optimal algorithm 'auto' because the current behavior is also kind of automatic. Maybe a better name would be method='optimal' ?
So we will have four different methods:
'sparse': always uses the sparse implementation
'dense': always uses the dense implementation
'input_type': choses the algorithm that matches the _type_ of the C matrix
'optimal': choses the best algorithm according to the sparsity of the C matrix. 

I'm also not certain how to handle the type of the returned T matrix. Right now the type of the T matrix is always identical to the type of the C matrix. For options='sparse' of 'dense', what should be the type of the T matrix? Should method='sparse' always return a sparse matrix, even if the C matrix was dense?
Maybe @trendelkampschroer has some opinions about this?

Or should we introduce an additional option to let the user define the return type? 

>    Your description of when you switch is unclear. First it's dimension**2 rather than dimension that is used. Then the bottomline is basicly that you use sparse calculation if less than 1/3 of the elements are nonzero, and dense otherwise. So I would write it this way.

Yes that is a mistake in the description, it's dimension**2. I will rewrite the documentation.

franknoe:
Am 18/06/15 um 18:44 schrieb fabian-paul:

> ```
> I find the parameter auto_sparse=True unclear. It is really meant
> to control the computation method and has nothing to do with the
> data type. For selection different computation methods, I
> generally suggest to use the parameter name "method", which could
> here take values 'auto' or 'dense' or 'sparse'. So the default
> would be method='auto'
> ```
> 
> I think there should also be a value of method='input_type' that 
> reproduces the current behavior of the method. Right now (before my 
> changes), tmatrix() uses the sparse algorithm if the input type is 
> sparse and the dense algorithm if the input type is dense. It might be 
> useful to provide this behavior for the user who needs it.
> 
> I don't think this is necessary. Data type and computation method are in 
> principle two different things. The current implementations use the 
> convention
>       input type = computation method = output type
> I would definitely keep the convention
>      input type = output type
> in order to avoid confusion, but there is no reason why we need to 
> enforce input type = computation method. If the user wants that they can 
> still choose sparse depending on the type (it's a low-level API). 
> "sparse", "dense", "auto" are sufficient.
> 
> Then I also would not name the method of choosing the optimal 
> algorithm 'auto' because the current behavior is also kind of 
> automatic. Maybe a better name would be method='optimal' ?
> So we will have four different methods:
> 'sparse': always uses the sparse implementation
> 'dense': always uses the dense implementation
> 'input_type': choses the algorithm that matches the /type/ of the C matrix
> 'optimal': choses the best algorithm according to the sparsity of the 
> C matrix.
> 
> I'm also not certain how to handle the type of the returned T matrix. 
> Right now the type of the T matrix is always identical to the type of 
> the C matrix. For options='sparse' of 'dense', what should be the type 
> of the T matrix? Should method='sparse' always return a sparse matrix, 
> even if the C matrix was dense?
> 
> No, see above. If you internally switch from dense to sparse to do a 
> sparse computation, you have to switch back to dense before returning.
> 
> Maybe @trendelkampschroer https://github.com/trendelkampschroer has 
> some opinions about this?
> 
> Or should we introduce an additional option to let the user define the 
> return type?
> 
> No, too complicated.
> 
> ```
> Your description of when you switch is unclear. First it's
> dimension**2 rather than dimension that is used. Then the
> bottomline is basicly that you use sparse calculation if less than
> 1/3 of the elements are nonzero, and dense otherwise. So I would
> write it this way.
> ```
> 
> Yes that is a mistake in the description, it's dimension**2. I will 
> rewrite the documentation.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/363#issuecomment-113213963.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

fabian-paul:
Ok.

fabian-paul:
It think, that's it.

franknoe:
Yes. Thank you!!!

PR #365, opened 2015-06-17
[pyemma.config] simplified access to config via attributes 

marscher:
eg. 
`pyemma.config.show_progress_bars = False`


franknoe:
ok

PR #370, opened 2015-06-19
Feature cossin

marscher:


franknoe:
Sorry, this feature is not complete:
- If cossin=True, you need to return the double dimension in the featurizer
- If cossin=True, you now have different features and thus need to change the describe() to contain cos and sin of that angle.
- The docstring is a bit misleading because one could think that you get (cos(a1), sin(a1), cos(a2), sin(a2),...) rather than (cos(a1), cos(a2), ... sin(a1), sin(a2), ...). I actually like the first version better, if it can be efficiently done.

marscher:
It is possible to achieve our third point with numpy.dstack and a reshape.
Also adopted the description strings and dimensions accordingly.

franknoe:
Still not quite right. describe() returns one label string _per feature_. cos(a) and sin(a) are two features, so they need to generate two strings. The number of strings returned by describe() should always be identical to dimension() (this could be a unit test).

marscher:
Paretos principle applies to this request now :) will change it.

franknoe:
looks good. Thanks @marscher

PR #379, opened 2015-06-25
[TICA] variable self._skipped_trajs stores skipped trajs and 'skipped traje…

gph82:
…ctory' warnings stop after 100 warnings

Otherwise the notebook might get too cluttered (or even crash) when thousands of warnings are issued.

PR #381, opened 2015-06-25
[cluster_regspace] produce ONLY ONE max_centers warning

gph82:
Trying to reduce the verbosity level which kill the notebooks when having thousands of trajectories.


marscher:
didn't we agreed upon saving the indices of skipped trajectories in a list and log this once, the parameterization is finished?

gph82:
@marscher  I think you are talking about #379 ?

marscher:
You're right, sorry. But actually this warning should appear only one, because it should stop executing the loop producing this warning.

gph82:
as a matter of fact, it does not (which I was wondering why in the first place).

 Perhaps, you can take a look at it?

marscher:
will do. Thank you

PR #382, opened 2015-06-25
[coor/transformer|regspace] introduced exception type for early parametrization termination 

marscher:
The only usecase for this is currently regspace clustering, which terminates when max_centers has been reached before seeing all data.

At first stance I tried to add the add_data_finished variable to the outer loops, which caused strange side effects in kmeans (cluster centers were not being converted). This is due to some algorithms expect to be called with last_chunk=True once. So raising an exception to terminate all loops seems to be the simplest and robust solution.

Superseeds #381

### DIFF
--- a/pyemma/coordinates/clustering/regspace.py
+++ b/pyemma/coordinates/clustering/regspace.py
@@ -21,6 +21,7 @@
 # ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+from pyemma.util.exceptions import NotConvergedWarning
 
 '''
 Created on 26.01.2015
@@ -128,8 +129,6 @@ class RegularSpaceClustering(AbstractClustering):
                                self.metric, self._max_centers)
             # finished regularly
             if last_chunk:
-                self.clustercenters = np.array(self._clustercenters)
-                self.n_clusters = self.clustercenters.shape[0]
                 return True  # finished!
         except RuntimeError:
             msg = 'Maximum number of cluster centers reached.' \
@@ -140,12 +139,15 @@ class RegularSpaceClustering(AbstractClustering):
             # finished anyway, because we have no more space for clusters. Rest of trajectory has no effect
             self.clustercenters = np.array(self._clustercenters)
             self.n_clusters = self.clustercenters.shape[0]
-
-            return True
+            # TODO: pass amount of processed data
+            raise NotConvergedWarning
 
         return False
 
     def _param_finish(self):
+        self.clustercenters = np.array(self._clustercenters)
+        self.n_clusters = self.clustercenters.shape[0]
+
         if len(self._clustercenters) == 1:
             self._logger.warning('Have found only one center according to '
                                  'minimum distance requirement of %f' % self.dmin)
--- a/pyemma/coordinates/data/feature_reader.py
+++ b/pyemma/coordinates/data/feature_reader.py
@@ -143,6 +143,15 @@ class FeatureReader(ReaderInterface):
         return patches.iterload(filename, chunk=self.chunksize,
                                 top=self.topfile, skip=skip, stride=stride)
 
+    def _close(self):
+        try:
+            if self._mditer:
+                self._mditer.close()
+            if self._mditer2:
+                self._mditer2.close()
+        except:
+            self._logger.exception("something went wrong closing file handles")
+
     def _reset(self, stride=1):
         """
         resets the chunk reader
@@ -185,9 +194,8 @@ class FeatureReader(ReaderInterface):
             if __debug__:
                 self._logger.debug('closing current trajectory "%s"'
                                    % self.trajfiles[self._itraj])
-            self._mditer.close()
-            if self._curr_lag != 0:
-                self._mditer2.close()
+            self._close()
+
             self._t = 0
             self._itraj += 1
             self._mditer = self._create_iter(self.trajfiles[self._itraj], stride=stride)
--- a/pyemma/coordinates/data/numpy_filereader.py
+++ b/pyemma/coordinates/data/numpy_filereader.py
@@ -73,7 +73,7 @@ class NumPyFileReader(ReaderInterface):
     def _reset(self, stride=1):
         self._t = 0
         self._itraj = 0
-        self._close_file()
+        self._close()
 
     def describe(self):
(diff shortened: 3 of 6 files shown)


franknoe:
- Do we really wanna exit in that case? One might loose a lot of computation time. I thought we would just stop clustering, issue a warning, and use the existing clusters for assignment at that point.
- If we use an exception, what about something like NotConvergedException (or existing version of that, which I guess must exist in numpy or scipy)? It's really the same sort of thing as not being able to converge an iteration. We could also use that for other problems. For example, is there such a thing in the maximum-likelihood transition matrix estimators, or do we just generate a warning if not converged there?

gph82:
On the first point: 
- I would agree if the warning also produces something like an estimate of how much data has been covered before exiting. It's not the same to exit after 80% of the data has been used or 10%. This warning could happen at the level of the assignment, as well. Those points that lie farther than dmin from any centers are precisely the data that was not visited when clustering. 

On the second point: +1 on all. 
- However you said something about changing a bit the design of the estimators, so  I'm not sure.

marscher:
@franknoe It is a "clean" exit, so one can use the clusters computed so far. That way you do not loose any information, but only terminate faster. The issue existed so far, was that even if max_clusters has been reached, the parametrization was continued (which was not intended). That is why @gph82 opened a PR to silence the warning "max_clusters has been reached", which showed up for every chunk after this checkpoint.

I agree to use a more general available exception and include the amount of data used for finding the centers.

franknoe:
Not sure if I understand what clean exit means. If you have a command 
such as (qualitatively)

``` python
rs = cluster_regspace(dmin=1.0)
rs.set_input(X)
rs.parametrize()
```

then I can imagine that you still can access the existing clusters. But 
you'll have to do the assignment yourself, because the parametrization 
will not complete. But in this command:

``` python
rs = cluster_regspace(X, dmin=1.0)
```

I don't think that rs will be available at all, so we can access it. Or 
am I wrong?

Am 26/06/15 um 14:02 schrieb Martin K. Scherer:

> @franknoe https://github.com/franknoe It is a "clean" exit, so one 
> can use the clusters computed so far. That way you do not loose any 
> information, but only terminate faster. The issue existed so far, was 
> that even if max_clusters has been reached, the parametrization was 
> continued (which was not intended). That is why @gph82 
> https://github.com/gph82 opened a PR to silence the warning 
> "max_clusters has been reached", which showed up for every chunk after 
> this checkpoint.
> 
> I agree to use a more general available exception and include the 
> amount of data used for finding the centers.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/382#issuecomment-115656111.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
Questions, where in _param_add_data there is an assignment being
performed? I do not see it :)

franknoe:
OK, so maybe I have missed that this has changed, but the second issue 
is still true, right? If you directly initialize cluster_regspace with 
data and it will throw an exception, you won't get anything.

Am 26/06/15 um 14:16 schrieb Martin K. Scherer:

> Questions, where in _param_add_data there is an assignment being
> performed? I do not see it :)
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/382#issuecomment-115658294.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
Am 26.06.2015 um 18:03 schrieb Frank Noe:

> OK, so maybe I have missed that this has changed, but the second issue
> is still true, right? If you directly initialize cluster_regspace with
> data and it will throw an exception, you won't get anything.
> No that's not an issue. The assignment step has nothing to do with the
> parametrization. This is being computed, if one accesses dtrajs
> attribute the first time.

This is common in all clustering algos. So maybe we should change that.

franknoe:
I think we are talking about different things. My argument is: when I am 
using ipython and

```
python rs = cluster_regspace(X, dmin=1.0)
```

exits with an exeption, then python interpreter exits berfore the 
variable rs is assigned, so the regspace object is unavailble to the 
user. Ergo they can't go on.

Am 26/06/15 um 19:07 schrieb Martin K. Scherer:

> Am 26.06.2015 um 18:03 schrieb Frank Noe:
> 
> > OK, so maybe I have missed that this has changed, but the second issue
> > is still true, right? If you directly initialize cluster_regspace with
> > data and it will throw an exception, you won't get anything.
> > No that's not an issue. The assignment step has nothing to do with the
> > parametrization. This is being computed, if one accesses dtrajs
> > attribute the first time.
> 
> This is common in all clustering algos. So maybe we should change that.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/382#issuecomment-115767513.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
No, because we handle the exception internally.
rs will be assigned and is perfectly valid.

franknoe:
Ah, now I understand. Thanks.

Am 27/06/15 um 03:26 schrieb Martin K. Scherer:

> No, because we handle the exception internally.
> rs will be assigned and is perfectly valid.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/382#issuecomment-115933801.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

fabian-paul:
This also brings up an interesting topic: exception handling in the pipeline. What happens if some error occurs in the later stages of a pipeline? I think the preceding stages in the pipeline have to be informed as well about the occurrence of the error. This is necessary for releasing resources like open file handles and buffers. 
This case is also related to the premature end of processing in the case of regspatial clustering. Here too the pulling from the pipeline is terminated prematurely and the file handles have to be closed. So with that similarity of these cases in mind, using an exception in regspatial for premature termination sounds like a good idea.
Then we actually have to implement the closing of files. I suggest that every transformer should have a .close() method which among other things calls the data_producer's close() method. That's how the information about the mandated termination could be propagated up the pipeline. 
Then thowing an exception (or certain kinds of exceptions) in a transformer should trigger the transformer's close() method.
What do you think?

marscher:
I'll think the close method would be very useful. I can implement this in this PR, so it is more complete.

marscher:
If there are no further comments, I'm going to merge this soon.

franknoe:
ok

Am 02/07/15 um 11:11 schrieb Martin K. Scherer:

> If there are no further comments, I'm going to merge this soon.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/382#issuecomment-117971529.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

PR #383, opened 2015-06-26
[coordinates/tica] fixed computaion of mean when force_eigenvalues_le_one is true

fabian-paul:
Because there are already some users of that feature (@gph82) this PR implements the correct way to compute the mean. (So as to remove the process that corresponds to eigenvalue one.)
I've extended the TICA test. It now checks the mean of the ICs whether it is really zero.
@gph: could you please check my changes with some of your data?

(This issue has nothing to do with order in which lag and stride are applied.)

### DIFF
--- a/pyemma/coordinates/tests/test_featurereader_and_tica_projection.py
+++ b/pyemma/coordinates/tests/test_featurereader_and_tica_projection.py
@@ -95,7 +95,7 @@ class TestFeatureReaderAndTICAProjection(unittest.TestCase):
         
     def test_covariances_and_eigenvalues(self):
         reader = FeatureReader(self.trajnames, self.temppdb)
-        trans = TICA(lag=1, output_dimension=self.dim, force_eigenvalues_le_one=True)
+        trans = TICA(lag=1, dim=self.dim, force_eigenvalues_le_one=True)
         trans.data_producer = reader
         for tau in [1, 10, 100, 1000, 2000]:
             log.info('number of trajectories reported by tica %d' % trans.number_of_trajectories())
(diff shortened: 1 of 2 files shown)


gph82:
Hi @fabian-paul , what exactly is the fix fixin?. I know it's "correcting the way to compute the mean", but what was wrong the previous computation?

fabian-paul:
In the variational principle, there is the stationary process that corresponds to the eigenvalue one and right eigenvector one (eigenfunction_1(x)=1). This process is removed in TICA by "subtracting the mean". As Feliks has shown "subtracting the mean" actually means subtracting the projection onto the the stationary right eigenvector such that all remaining processes are uncorrelated to the stationary process. For the old way of counting this is identical to subtracting the mean.
But because I have (slightly) changed the way how the correlation matrix is computed, this affects also the the way how the mean should be removed.
More mathematically speaking "removing the mean" is the transformation:
x -> x - cov(x,1)
I've just adapted the cov()-function that is used in the mean calculation to the cov-function that is used to compute the covariances.

gph82:
so in the old implementation, we were not exactly removing the stationary vector that would correspond to the computed eigenvectors, but the stationary vector that corresponds to the old way of computing the covariance matrices.

fabian-paul:
yes

gph82:
ok, I was not observing any strange behavior, but I'll check with the new impl once it's merged.

fabian-paul:
just to be clear: the right stationary eigenvector doesn't have to be computed, because we know it is always one. The crucial idea is that the mean is actually the covariance of the data with the one-function.

PR #390, opened 2015-07-03
[frames_from_file] topology argument more flexible

gph82:
Now topologies can be parsed not only in form of filenames but actually as true mdtraj.Topology or mdtraj.Trajectory objects. This avoids reading the pdbfile every frames_from_file gets called.

In the "worst-case-scenario" for save_traj and save_trajs (eg. a sample of 50 frames composed by 1 frame of 50 different files), the speedup is of about 1.5

### DIFF
--- a/pyemma/coordinates/api.py
+++ b/pyemma/coordinates/api.py
@@ -53,6 +53,9 @@ from pyemma.coordinates.clustering.assign import AssignCenters as _AssignCenters
 # stat
 from pyemma.coordinates.util.stat import histogram
 
+# types
+from mdtraj import Topology as _Topology, Trajectory as _Trajectory
+
 import numpy as _np
 import itertools as _itertools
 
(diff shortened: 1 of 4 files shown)


franknoe:
Good! Can we use the same argname like mdtraj? I think they use top

franknoe:
P.S.: You have to consider for each such change that you break people's code because you have changed the name of the keyword. In this case I think it's ok to do that because it's a minor option, but generally we should avoid that.

gph82:
this got merged while I was PR'in some other things, I think I am going to rebase and update the #PR391

franknoe:
Ah sorry, I thought this was finished.

Am 03/07/15 um 19:37 schrieb Guillermo Pérez-Hernández:

> this got merged while I was PR'in some other things, I think I am 
> going to rebase and update the #PR391
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/390#issuecomment-118398440.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
and it was, it's that I could've waited a bit more, pull the merged changes and develop #391 on that, now I'm lost between devel and dev_release_1.3 the mergetool jungle

PR #391, opened 2015-07-03
[frames_from_file] option to preallocate trajectory

gph82:
This again, like #390, aims at speeding up the generation of trajectory objects with frames coming from different files and in arbitrary order. 

The actual change happens in frames_from_file, but it propagates up to the api calls of save_trajs and save_traj through the parameter copy_not_join.

Speedup in the normal usecase of frames_from_file is 1.4. 

Check out copy_traj_attributes in reader_utils.py to see what happens.

As for the API calls, perhaps there are better ideas, variable names etc,  but this works for now. 

Perhaps it is time to implement **kwargs of frames from file and simplify these calls?


gph82:
@franknoe , can you take a look at this see what you think?

gph82:
I think I solved the merge conflict

gph82:
please don't merge yet, still working on some improvements with @marscher

franknoe:
what's the stage of this PR? Should it be merged, closed or still on hold?

gph82:
Still on hold.

gph82:
Please hold

gph82:
After discussion with @clonker about his benchmarks of #458, we think that it's best if we combine both our approaches in frames_from_file.py. There, his random access (PR #458) will be used for everything but .xtcs and trr (for which we will keep use the stuff in this PR). This Fallunterscheidung will be invisible for the user. Hence, the api calls of save_traj(s) DO NOT CHANGE (contradicting the comment that opens this PR)

If (and when) the tests pass, you can merge

gph82:
Then @clonker can take it from there.

PR #393, opened 2015-07-06
[msm/flux] Pathway decomposition uses floating point tolerance when c…

trendelkampschroer:
…hecking the termination condition (current flux >= fraction \* total flux) in the main iteration

### DIFF
--- a/pyemma/msm/flux/sparse/pathways.py
+++ b/pyemma/msm/flux/sparse/pathways.py
@@ -235,7 +235,7 @@ def remove_path(F, path):
     return F
 
 
-def pathways(F, A, B, fraction=1.0, maxiter=1000):
+def pathways(F, A, B, fraction=1.0, maxiter=1000, tol=1e-14):
     r"""Decompose flux network into dominant reaction paths.
 
     Parameters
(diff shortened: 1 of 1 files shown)


franknoe:
What is `tol=1e-14` for?

trendelkampschroer:
If decomposition is requested with fraction=1.0 the iteration has to successfully remove all reaction paths from the netflux network. This removal can introduce numerical errors since the flux matrix has to be updated after the detection of every path. I have observed that the algorithm can fail to discover pathways with very small capacity values at the end of the iteration,  i.e. the capacity of all previously found paths CF and the total capacity (or total flux) TF are almost equal/ fraction=1.0-tol. 

In order to prevent the algorithm from trying to detect paths (in a flux matrix possibly containing only numerical noise) and failing I have modified the termination criterion for the main iteration accordingly.

franknoe:
Ah OK. Please document the argument then. I'll merge after the update.
I guess the high-level API (ReactiveFlux) is not affected by your change?

trendelkampschroer:
The high level api is not affected by this change.

Will update the documentation and send a new pull request.

PR #406, opened 2015-07-09
[ImpliedTimescales] use joblib for parallel model estimation.

marscher:


franknoe:
I think we forgot about that. Is that still up to date?

marscher:
I rewrote this on top of the new estimator class, so it is up to date. The estimation of tmatrices and diagonalization has been parallelized.

PR #410, opened 2015-07-09
Estimator tests

franknoe:
- Started a standard approach to include test datasets into emma. These are now under pyemma.datasets, and are thus not only available for software developers to write unit tests, but for everyone using the package.
- Added several integration tests for ImpliedTimescales using different underlying Estimators (MaximumLikehoodMSM, HMSM and BayesianMSM, HMSM)
- Improved msm.api docs and added examples.


marscher:
appveyor seems to work again.. 

With #411 in mind, I'll suguest using npy (or compressed ones) files for test-data.

franknoe:
Agreed, I will replace the data files with binaries.

PR #413, opened 2015-07-10
Remove msm package

marscher:
This PR removes the pyemma.msm package despite the models, estimators and utils packages and maps them to msmtools, which is already available in conda and pypi. The mapping forbids relative imports in msmtools, which only occurred  in estimation/dense/tmatrix.py. The fix for this has not yet been released.

Also fixed the documentation references, which were not refactored.


marscher:
The import hook is only necessary for nosetests, since the override the default import behaviour (they import from paths/directories, instead of respecting the imports in msm/**init**.py). As soon we figure out how to avoid this, we can safely remove the hook.

trendelkampschroer:
Can you check if the --no-path-adjustment flag to nose is able to solve the import issues.

http://nose.readthedocs.org/en/latest/usage.html#cmdoption--no-path-adjustment

marscher:
unfortunately adding this option to nose does not change anything

trendelkampschroer:
What a pity, maybe we'll find something clever to solve this in the mean time we should stick to the solution you proposed.

marscher:
which would require a new msmtools release. Is there anything to include before doing that? The floating counts fix PR from @fabian-paul opened against Pyemma should be merged in msmtools.

trendelkampschroer:
I am working on stuff but it might take a few days until I have that ready and can take a look at the fix Fabian proposed. 

I don't have a problem with making two minor releases in short succession, do whatever you think is best so that development for pyemma doesn't get blocked by this. 

@fabian-paul: Can you do your pull-request against msmtools so that I can see what is happening there

marscher:
everything is fine, except the testing coverage... which is somehow confusing, since we deleted a lot of code.

franknoe:
I'm not surprised. The test coverage was higher in the msm low-level 
part than in the rest of the code, so the average coverage went down.

Am 14/07/15 um 20:10 schrieb Martin K. Scherer:

> everything is define, except the testing coverage... which is somehow 
> confusing, since we deleted a lot of code.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/413#issuecomment-121326978.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
sure now it gets revealed :+1:

marscher:
In my opinion this one is in a merge-able state now.

trendelkampschroer:
Looks good to me.

PR #432, opened 2015-07-20
[featurizer.py] minimum distance between residues and minimal distances between groups of atoms

gph82:
As usual, this is more a basis for discussion/refactoring renaming than an PR
These two new features (which are very similar) are presented separated to keep the input simple:
- add_residue_mindist only operates with residue based indexing, and is just an interface to http://mdtraj.org/latest/api/generated/mdtraj.compute_contacts.html. I
- add_group_mindist is very flexible, and allows the user to define arbitrary atom groups via atom indices. The user can ask mindists among all groups or only some of them. Eg., if 4 groups have been defined, perhaps the user wants mindist_0-1 and mindist_2-3, but not mindist_0_3 etc. Think of two separate salt bridges or two separate ionic-locks in different parts of the molecule. If given only one group, add_group_mindist automatically assumes the minimal distance within that group is wanted (not clear about the use-case, but still)
- _parse_groupwise_input mimics what _parse_pairwise_input does, but with a group-type of input. Perhaps in the future more features will be defined among groups and this parser/input-checker is useful. Already discussed with @marscher  the possibility of outsourcing it to some other util file, or even to the feature object itself. I really do not have an opinion about that.

### DIFF
--- a/pyemma/coordinates/data/featurizer.py
+++ b/pyemma/coordinates/data/featurizer.py
@@ -354,10 +354,11 @@ class InverseDistanceFeature(DistanceFeature):
 
 class ResidueMinDistanceFeature(DistanceFeature):
 
-    def __init__(self, top, contacts, scheme, ignore_nonprotein):
+    def __init__(self, top, contacts, scheme, ignore_nonprotein, threshold):
         self.top = top
         self.contacts = contacts
         self.scheme = scheme
+        self.threshold = threshold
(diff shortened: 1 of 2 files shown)


gph82:
Hi, please don't merge yet, I'll do some more coding later.

franknoe:
Thanks a lot Guille. Looks good on first sight, but I guess we have to work with it in order to track down possible performance issues.

PR #451, opened 2015-07-27
Test msm sparse

trendelkampschroer:
Extended unit test for MSM object to also cover sparse MSMs


franknoe:
Thanks a lot Benjamin!

That Appveyor problem occurs all the time. Is that fixed now by the doctest fixes?

PR #458, opened 2015-07-28
sorted random access of frames and mini-batch kmeans

clonker:
This PR is from my side not quite finished, but I thought it would be a good idea to create it in order to have some discussion of the open points available.

It includes random access of frames by passing a 2xN array into the stride argument. This array must be sorted, otherwise an error is risen.
It also includes the mini-batch kmeans algorithm. Currently, all data gets aggregated and then the cluster centers get updated. This behaviour can of course be changed to a streaming one, but at least for the kmeans++ initialisation, all (sampled) data has to be available in memory anyways.

I am not sure, if the mini-batch-kmeans should be included into the coordinates api by an additional flag or by another function. I think the flag would be preferable but since the amount of the drawn frames has to be specified as well, one would need even another argument. On the other hand another api-function requires documentation which would be mostly duplicated. 

Further, i have changed the convergence criterion. Instead of looking if the centers changed, the kmeans cost function is evaluated. The problem with the centers is that each sample of the data causes some small oscillations. So if the sample size is low (like >50%) and the tolerance is small (like currently 1e-5), the algorithm might never converge. Evaluation of the cost function is in O(n_centers \* n_frames).
However, in the case of mini-batch kmeans, of course only the currently drawn sample gets evaluated, which again kind of defeats the purpose of this convergence criterion, but it might be a good-enough approximation.


franknoe:
Thanks clonker, this is a major contribution! I'll need some time to look through this. Expect my comments on thursday.

pyemma/coordinates/util/patches.py
```diff
@@ -120,3 +160,27 @@ def iterload(filename, chunk=100, **kwargs):
...
+    top = trajs[0].top
+    n_frames = sum(t.n_frames for t in trajs)
+
+    xyz = np.empty((n_frames, top.n_atoms, 3))
+    t = 0
+    for traj in trajs:
+        n = len(traj)
+        xyz[t:n+t] = traj.xyz
+        t += n
+        # TODO: what about time?
```
gph82, line 186: I think it would be useful to append other arguments (since they will be available), as done in reader_utils.py of  https://github.com/gph82/PyEMMA/commit/c179aa7b364d0d71750ecc6a51abe552ecd322e3/ costs nearly nothing and makes the frame complete

gph82:
Great! Where exactly are you accessing the trajectories with the nx2-index array, from a molecular file?

clonker:
In iterload (patches.py, beginning at line 108) i am loading the indices of the current trajectory (there it is only a Nx1 array). When the iterator gets created, a call to 

``` python
context.ra_indices_for_traj(itraj)
```

is made which reduces the Nx2 array to the Nx1 array for the corresponding trajectory. You can find the code for ra_indices_for_traj in transformer.py somewhere at the very top.

pyemma/coordinates/transform/transformer.py
```diff
@@ -35,15 +35,89 @@
...
+            raise ValueError("Currently only sorted arrays allowed for random access")
+
+    def ra_indices_for_traj(self, traj):
+        """
+        Gives the indices for a trajectory file (without changing the order within the trajectory itself).
+        :param traj:
+        :return:
+        """
+        assert not self.uniform_stride, "requested random access indices, but is in uniform stride mode"
+        return self._stride[self._stride[:, 0] == traj][:, 1] if traj in self.traj_keys else np.array([])
```
gph82, line 63: If I understand correctly, @clonker, this object has a method that takes the (already loaded) trajectory as an argument? Is traj a chunk, and the indices in "stride" are shifted accordingly? Or is traj the entire trajectory object and the indices in stride are absolute frame indices?

pyemma/coordinates/util/patches.py
clonker, line 186: I think so too. :-) Perhaps when your PR is merged, so we avoid code duplication.

pyemma/coordinates/transform/transformer.py
clonker, line 63: I guess the documentation is not quite clear about this method.. The context object holds the Nx2 array (self._stride). The "traj" argument is then merely an index, not the actual trajectory and the return value is an np.array of indices for that specific trajectory(-index). 

The indices in stride are absolute per trajectory, there is no overlap, they always begin at 0. I mean that if trajectory "0" has length 10, then you (currently) can't assign the trajectory 0 index 11 and mean the first frame of trajectory "1".

gph82, line 63: I know that most of the effort is in the algorithmic implementation of mini-batch kmeans, but I am very interested in how ultimately the frames are accessed from the file, and how that process can be speed up. This https://github.com/markovmodel/PyEMMA/blob/devel/pyemma/coordinates/data/frames_from_file.py is a bottleneck for the FAH dataset right now

marscher, line 63: Actually you should use the new stride semantics in eg. save_traj to
obtain the frames efficiently to avoid code duplication or
re-implementing the same thing, because I think @clonker has done this
the most (io) efficient way possible.

gph82, line 63: exactly my point, @marcher, but in order to use it this way I need to understand it first. Hence my question, where are the files being accessed?

pyemma/coordinates/api.py
```diff
@@ -288,7 +289,7 @@ def source(inp, features=None, top=None, chunk_size=100):
         # check: if single array, create a one-element list
         # check: do all arrays have compatible dimensions (*, N)? If not: raise ValueError.
         # create MemoryReader
-        reader = _DataInMemory(inp)
+        reader = _DataInMemory(inp, chunksize=chunk_size)
```
marscher: The default chunksize for DataInMemory is 5000, because 100 was too small for most cases (high python overhead). But actually you're right to pass the argument there, because otherwise the user has to make an extra call to set the chunksize.

pyemma/coordinates/clustering/kmeans.py
```diff
@@ -193,3 +185,107 @@ def _param_add_data(self, X, itraj, t, first_chunk, last_chunk_in_traj, last_chu
...
+                cc = kmeans_clustering.init_centers(self._in_memory_chunks,
+                                                    self.metric, self.n_clusters)
+                self._cluster_centers = [c for c in cc]
+
+    def _collect_data(self, X, first_chunk, stride):
+        # beginning - compute
+        if first_chunk:
+            self._t_total = 0
+            mem_req = int(1.0 / 1024 ** 2 * X[0, :].nbytes * self.n_frames_total(stride))
+            if mem_req > 10:
```
marscher: please increase this to 100mb

pyemma/coordinates/transform/transformer.py
clonker, line 63: I think the only way to speed things up is to stream through the file once (and if order needs to be preserved, have a mapping or something) and collect the data with seek and whence. In my code it is like this (leaving out chunk sizes):

``` python
# some initialization
x_prev = 0
curr_size = 0
# this loop groups together subsequent indices, so [1,2,3,5,6,8,9] becomes [1,2,3],[5,6],[8,9]
for k, g in groupby(enumerate(stride), lambda (a, b): a-b):
    # this here is the current group, i.e., [5,6]
    grouped_stride = map(itemgetter(1), g)
    # some offset because the trajectory begin has some special handling
    seek_offset = (1 if x_prev != 0 else 0)
    # this is how far we need to seek
    seek_to = grouped_stride[0] - x_prev - seek_offset
    # do the seek
    f.seek(seek_to, whence=1)
    # remember to which index we read the trajectory so we can calculate seek_to for the next group
    x_prev = grouped_stride[-1]
    # how many frames do we read
    group_size = len(grouped_stride)
    # read it!
    traj = _get_local_traj_object(atom_indices, extension, f, group_size, topology, **kwargs)
```

after having read the trajectories for the index groups, they get concatenated and yielded.

Of course it all depends on how efficient seek is implemented on the mdtraj side as well.

gph82, line 63: I ran some tests with a very similar implementation of get_frames... using seek, did not really see much of a speedup, but it might have been my implementation. We can dig it up and check. I all for getting rid of frames_from_file if there are better ways.

pyemma/coordinates/clustering/kmeans.py
```diff
@@ -193,3 +185,107 @@ def _param_add_data(self, X, itraj, t, first_chunk, last_chunk_in_traj, last_chu
...
+        for idx, traj_len in enumerate(self._traj_lengths):
+            traj_samples = int(math.floor(traj_len / float(self._total_length) * samples))
+            self._n_samples_traj[idx] = traj_samples
+            self._n_samples += traj_samples
+
+        self._random_access_stride = np.empty(shape=(self._n_samples, 2), dtype=int)
+
+        super(MiniBatchKmeansClustering, self)._param_init()
+
+        return 0, self._draw_mini_batch_sample()
```
marscher, line 258: Why do you return the sample here? We used the first return value in Transformer.parametrize for requesting time-lagged data and I can not see, that you have changed this code to accept multiple return values.

marscher, line 258: sorry, I didn't spot it in the first place. But a comment, why you are returning it at this place would be better to understand the code.

pyemma/coordinates/util/patches.py
marscher, line 186: The pr was requested to be on hold by @gph82, wasn't it? Is it mergeable now?

pyemma/coordinates/api.py
clonker: How do we deal with this, set chunk_size to None by default and then set it to 100 for anything but DataInMemory if not set otherwise?

pyemma/coordinates/clustering/kmeans.py
clonker, line 258: Actually, i have - in transformer.py lines 331 up to 339. This is necessary because i want to have a sampled version of the data before _param_add_data gets called for the first time.

pyemma/coordinates/transform/transformer.py
clonker, line 63: It just might be the mdtraj implementation as well (and therefore the trajectory file format).. i am happy to run some performance tests.

gph82, line 63: actually you can pretty easily, in your branch. Try to obtain two equal
trajectory objects, one with this sorted-random-access and once with
frames_from_file.py

On 28.07.2015 16:40, Moritz Hoffmann wrote:

> In pyemma/coordinates/transform/transformer.py
> https://github.com/markovmodel/PyEMMA/pull/458#discussion_r35654764:
> 
> > -            keys = stride[:, 0]
> > -            self._trajectory_keys, self._trajectory_lengths = np.unique(keys, return_counts=True)
> > -        else:
> > -            self._trajectory_keys = None
> > -        self._uniform_stride = TransformerIteratorContext.is_uniform_stride(stride)
> > -        if not self.uniform_stride and not self.is_stride_sorted():
> > -            raise ValueError("Currently only sorted arrays allowed for random access")
> >   +
> > -    def ra_indices_for_traj(self, traj):
> > -        """
> > -        Gives the indices for a trajectory file (without changing the order within the trajectory itself).
> > -        :param traj:
> > -        :return:
> > -        """
> > -        assert not self.uniform_stride, "requested random access indices, but is in uniform stride mode"
> > -        return self._stride[self._stride[:, 0] == traj][:, 1] if traj in self.traj_keys else np.array([])
> 
> It just might be the mdtraj implementation as well (and therefore the
> trajectory file format).. i am happy to run some performance tests.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/458/files#r35654764.

## 

Dr. Guillermo Pérez-Hernández
Freie Universität Berlin
Institute for Mathematics
Arnimallee 6
D-14195 Berlin
tel       0049 30 838 75776

## http://userpage.fu-berlin.de/gph82/

pyemma/coordinates/clustering/kmeans.py
clonker: done

pyemma/coordinates/clustering/kmeans.py
clonker, line 258: added a comment

pyemma/coordinates/api.py
marscher: Am 28.07.2015 um 16:35 schrieb Moritz Hoffmann:

pyemma/coordinates/transform/transformer.py
clonker: Alright, I have run some tests now. The speed **strongly** depends on the file format which is used. For the file formats that were faster i have randomly selected 10.000 samples from 50.000 frames and for the slower formats 1.000 from 10.000 frames.

1. 10.000 samples from 50.000 frames:

Format | Frames from files | Random access Stride
------ | --------------- | --------------------
h5 | 83 seconds | 21 seconds
dcd | 86 seconds | 3 seconds
binpos | 86 seconds | 2 seconds
nc | 86 seconds | 3 seconds

2. 1.000 samples from 10.000 frames:

Format | Frames from files | Random access Stride
------ | --------------- | --------------------
h5 | 1.76 seconds | 2.08 seconds
dcd | 1.62 seconds | 0.31 seconds
binpos | 1.57 seconds | 0.17 seconds
nc | 1.62 seconds | 0.33 seconds
xtc | 1.65 seconds | 23.58 seconds
trr | 1.73 seconds | 52.45 seconds

... as you can see, xtc and trr are quite bad formats for the 'seek-whence'-approach. I have found this issue rmcgibbo/mdtraj#336, which suggests that xtc's seek is not quite efficient.

clonker:
Alright, I have run some tests now. The speed **strongly** depends on the file format which is used. For the file formats that were faster i have randomly selected 10.000 samples from 50.000 frames and for the slower formats 1.000 from 10.000 frames.
1. 10.000 samples from 50.000 frames:

| Format | Frames from files | Random access Stride |
| --- | --- | --- |
| h5 | 83 seconds | 21 seconds |
| dcd | 86 seconds | 3 seconds |
| binpos | 86 seconds | 2 seconds |
| nc | 86 seconds | 3 seconds |
1. 1.000 samples from 10.000 frames:

| Format | Frames from files | Random access Stride |
| --- | --- | --- |
| h5 | 1.76 seconds | 2.08 seconds |
| dcd | 1.62 seconds | 0.31 seconds |
| binpos | 1.57 seconds | 0.17 seconds |
| nc | 1.62 seconds | 0.33 seconds |
| xtc | 1.65 seconds | 23.58 seconds |
| trr | 1.73 seconds | 52.45 seconds |

... as you can see, xtc and trr are quite bad formats for the 'seek-whence'-approach. I have found this issue mdtraj/mdtraj#336, which suggests that xtc's seek is not quite efficient.

gph82:
@clonker , thanks! Pretty useful! 

I ran all of my tests exclusively on .xtcs. 
Can I please ask you to check out https://github.com/gph82/PyEMMA/commit/c179aa7b364d0d71750ecc6a51abe552ecd322e3 which further optimizes frames_from_file.py (copy_not_join=True) and see how the numbers change?

marscher:
The problem with xtc seek is that it has to read every frame in between 
two desired positions. If one wants to avoid this problem, one has to 
create some sort of index first. The reader in Stallone java library did 
this to achieve good random access performance.

marscher:
maybe we should suggest mdtrajs people to use libxdrfile2: 
https://pythonhosted.org/MDAnalysis/documentation_pages/coordinates/XTC.html

clonker:
Of course! Your numbers changed for the 1000 samples more or less consistently to about 1.53 seconds and for the 10.000 samples they dropped to about 70 seconds.

But you should keep in mind that I currently only allow sorted access -- so there will be some cutback in the access times once unsorted access is allowed as well.

clonker:
This libxdrfile2 looks good, the page also contains a remark about the trr-files, which apparently require an index to be built for fast random access as well.

gph82:
BTW, @clonker , these test take into account re-ordering the indexes, right? The randomly drawn frames get sorted for retrieval, than re-shuffled (to their original random order) for output, right? Are you also measuring this?

marscher:
Unfortunately libxdrfile2 is GPLv2 which can not be relicensed under 
LGPL, if I'm not mistaken.

franknoe:
exactly, that's what I meant.

Am 28/07/15 um 20:53 schrieb Martin K. Scherer:

> The problem with xtc seek is that it has to read every frame in between
> two desired positions. If one wants to avoid this problem, one has to
> create some sort of index first. The reader in Stallone java library did
> this to achieve good random access performance.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/458#issuecomment-125716556.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

clonker:
@gph82: The input data i used for the tests was sorted so far, since the api random-access implementation does not allow otherwise. So if you mean by taking into account that i was using shuffled data, no, but if you meant if i was measuring the time your algorithm took to sort, fetch and re-shuffle (even though there was nothing to sort), yes.
But funnily enough the file-reader random-access implementation already can do it (even though i didnt expect it to; by jumping around in the file, likely less efficient than sorting and then reshuffeling). Here are the test results for 10.000 samples out of 50.000 available frames:

| Format | Frames from files | Random access stride |
| --- | --- | --- |
| h5 | 69.5 seconds | 26.1 seconds |
| dcd | 70.7 seconds | 39.0 seconds |
| binpos | 70.75 seconds | 1.86 seconds |
| nc | 69.6 seconds | 3.4 seconds |

I took your improved version with copy_not_join=True.

marscher:
merged in dev_release and then to devel

PR #461, opened 2015-07-28
[tica] added implied timescales attribute

gph82:
I think this is nice

### DIFF
--- a/pyemma/coordinates/transform/tica.py
+++ b/pyemma/coordinates/transform/tica.py
@@ -210,6 +210,8 @@ class TICA(Transformer):
         self._N_mean = 0
         self._N_cov = 0
         self._N_cov_tau = 0
+        # create mean array and covariance matrices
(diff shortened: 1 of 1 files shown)


franknoe:
There is a minus missing. Your timescales are negative

gph82:
OMG, that's right, even in the test! (which BTW I'm renaming because it had a wrong name)

gph82:
Just updated the PR after merging with dev_release_1.3 We should be okay

gph82:
scratch that. I'll take care of it later

gph82:
I think you have to revert this, @marscher

marscher:
Am 29.07.2015 um 19:15 schrieb Guillermo Pérez-Hernández:

gph82:
tests are failing...for reasons

marscher:
Will fix it on the branch directly. Thanks!

Issue #331, opened 2015-05-27
[coordinates.tica] include input mean option


gph82:
If want to compute TICA at two (or more) different taus, it should be possible to skip the first pass of 
TICA once the mean is already known.

franknoe:
And C0, which is more expensive.

Is it still true that we can skip the mean and C0 with the way of 
counting that Fabian has suggested? I think there both the mean and C0 
might depend on tau, but I don't remember exactly.

Am 27/05/15 um 11:44 schrieb Guillermo Pérez-Hernández:

> If want to compute TICA at two (or more) different taus, it should be 
> possible to skip the first pass of
> TICA once the mean is already known.
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/markovmodel/PyEMMA/issues/331.

## 

---

Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

## Mail:  Arnimallee 6, 14195 Berlin, Germany

gph82:
Yes, the less we recompute, the better. 

However, before we get into implementing, testing etc the @fabian-paul method, just skipping the first pass should be easy to implement. I would prioritize this now. It's a huge timesaver for the FAH dataset

franknoe:
I suggest to do this in your own fork if you need it now.

For the devel branch and future releases, I would like to sort out the 
basic questions to have something stable that we won't change again in 
the subsequent version. Questions include:
- The @fabian-paul method is already (optionally) used in pyEMMA, so if 
  it's means and C0 are tau-dependent we must exclude that both options 
  are used. Also we should discuss whether Fabian's counting method is 
  what we want to stick with and then make it the standard option.
- I think it is worthwhile considering to build separate estimators for 
  correlation matrices and means and have TICA use them. We will also need 
  them for other estimators such as the variational approach. But this is 
  a model decision we need to discuss and evaluate.

Am 27/05/15 um 11:49 schrieb Guillermo Pérez-Hernández:

> Yes, the less we recompute, the better.
> 
> However, before we get into implementing, testing etc the @fabian-paul 
> https://github.com/fabian-paul method, just skipping the first pass 
> should be easy to implement. I would prioritize this now. It's a huge 
> timesaver for the FAH dataset
> 
> —
> Reply to this email directly or view it on GitHub 
[...]

PR #463, opened 2015-07-28
[coor/pca|tica] added optional mean argument and skip first pass, if provided

### LINKED ISSUE #62: add method to store calculated mean and optionally pass it to ctor of PCA/TICA

marscher:
I introduced a new a new exception type (SkipPassException) to break the inner loops in Transformer.parametrize. In this case we can not simply return True in param_add_data, because this would terminate the whole parametrization routine.

Fixes #62, #331.


gph82:
This is great! 

What about the discussed (#331) lag-dependency of the mean? Should we issue a warning?

marscher:
added the warning

PR #465, opened 2015-07-29
[add_residue_mindist] gets residue_indices_2 argument

gph82:
I think this is something like a feature request from @nsplattner . This extends the input mode for 
add_residue_mindist to make it easier to map situations where ligand-substrate are clearly defined. 

I'm aware we're not that convinced about the name of this feature, but that is another discussion. 

Please, @nsplattner , if this is not exactly what would help you most for the notebooks, let me know.


nsplattner:
O.k., thanks! I will look into this as soon as I have time.

Issue #474, opened 2015-08-02
Please check phi/psi featurizer in 1.2.2 release


franknoe:
In this notebook:

```
http://www.emma-project.org/latest/generated/md2msm_penta_peptide.html
```

the featurizer description that shows up is this:

```
['COS(PHI 0 LEU 2 )',
 'SIN(PHI 0 LEU 2 )',
 'COS(PHI 0 ALA 3 )',
 'SIN(PHI 0 ALA 3 )',
 'COS(PHI 0 LEU 4 )',
 'SIN(PHI 0 LEU 4 )',
 'COS(PHI 0 LEU 5 )',
 'SIN(PHI 0 LEU 5 )',
 'COS(PHI 0 TRP 1 )',
 'SIN(PHI 0 TRP 1 )',
 'COS(PHI 0 LEU 2 )',
 'SIN(PHI 0 LEU 2 )',
 'COS(PHI 0 ALA 3 )',
 'SIN(PHI 0 ALA 3 )',
 'COS(PHI 0 LEU 4 )',
 'SIN(PHI 0 LEU 4 )']
```

which is wrong because PSI angles are missing. I don't think this is an issue in dev_release_1.3, so be careful when merging because I don't know if this was a problem not fixed before or a problem introduced by some redundancy checking in 1.2.2.

Please check and write unit tests in such a way that for all test cases the dimension of the featurizer and the number of features and the number of description labels are the same.

gph82:
Hi Frank, do you mean missing as in "the dimensionality of the transformer" is half the length of the list?

gph82:
Oh, sorry, I see now in the notebook number of dimension =  16...so what do you mean exactly?

gph82:
Oh man, i need a coffee. Scratch all of my comments. This got fixed here #434

franknoe:
Thanks. So we don't have it in 1.2.2. - that's a bit annoying because 
the Tutorial notebook shows up wrong.

In any case, in case if we don't have that yet, please make sure that 
feature dimensions _and_ consistenty between dimension() and 
len(describe()) are unit-tested in all featurizer unit tests. In this 
case we expect 4_nres dimensions when cossin=True, and 2_nres dimensions 
otherwise.

Am 02/08/15 um 17:56 schrieb Guillermo Pérez-Hernández:

> Oh man, i need a coffee. Scratch all of my comments. This got fixed 
> here #434 https://github.com/markovmodel/PyEMMA/pull/434
> 
> —
[...]

PR #475, opened 2015-08-02
[featurizer] tests check for len of description vs dimension

gph82:
Suggested in #474


franknoe:
Still failling the windows file test. Please fix that next week.

Thanks Guille!

PR #476, opened 2015-08-03
Synched devel with dev rel 1.3

marscher:
merged last rather huge PR on dev_release_1.3 and resolved merge conflicts with devel. @clonker can you please check, why these test failures in coordinates package occur? Maybe it is related to some merge conflict, but I was not able to spot it now.
It occurs in accumulation of data in kmeans

```
   self._collect_data(X, first_chunk, stride)
  "clustering/kmeans.py", line 225, in _collect_data
    self._in_memory_chunks[self._t_total:self._t_total + len(X)] = X[:]
ValueError: could not broadcast input array from shape (37,100) into shape (0,100)
```

Another issue is: (due the recently introduced option of passing means into TICA, but actually it does also fail, if this option is not used at all).

```
  File "tica.py", line 304, in _param_add_data
    self._N_cov_tau += 2.0 * np.shape(Y)[0]
IndexError: tuple index out of range
```

@fabian-paul can you please have a look?

The failures in msm package are due to a not yet merged PR in bhmm (call back function). @franknoe would be kind if you merge this soon, since it is blocker right now. The change is trivial.

--> Closed. Check out #477


franknoe:
Sorry, it doesn't make sense to submit a PR with tons of failures. 
One of the main problems here is that Benjamin has accidentally created his sparse MSM test against devel instead of dev_release_1.3 which creates a ton of conflicts. I'll close this for now and have a closer look.

PR #478, opened 2015-08-03
[featurizer.pairs] opt arg excluded_neighbors added

gph82:
This costs nothing, breaks nothing, makes pairs() more flexible and is compatible with existing code.

### DIFF
--- a/pyemma/coordinates/data/featurizer.py
+++ b/pyemma/coordinates/data/featurizer.py
@@ -808,23 +808,31 @@ class MDFeaturizer(object):
         return self.topology.select("mass >= 2")
 
     @staticmethod
-    def pairs(sel):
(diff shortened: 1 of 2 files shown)


franknoe:
Thanks, very good idea! Just two little requests:
- please now merge against devel not dev_release_1.3 (this explains the appveyor error). dev_release_1.3 is dead, long live dev_release_1.3. I will remove the branch
- please don't exclude anything on default, i.e. the default should be set such all N*(N-1)/2 pairs are used. I'm not sure if this is 0 or 1.

gph82:
"please don't exclude anything on default, i.e. the default should be set such all N*(N-1)/2 pairs are used. I'm not sure if this is 0 or 1."
This would be 0. Might conflict with existing code. I took exclude_neighbors=2 because that was the default value

PR #479, opened 2015-08-03
[featurizer.pairs] opt arg excluded_neighbors added, default behaviour changed

gph82:
This closes #478. @franknoe , if I set default exclusions=0 tests break. Might be that people are already using it that way. I'd suggest we leave it like this for now. The entire "pairwise_parse" area will be a bit of construction site once I have more time (hTICA needs very good handling of this)

### DIFF
--- a/pyemma/coordinates/data/featurizer.py
+++ b/pyemma/coordinates/data/featurizer.py
@@ -907,8 +907,6 @@ class MDFeaturizer(object):
 
                 iterable of integers (either list or ndarray(n, dtype=int)):
                     indices (not pairs of indices) of the atoms between which the distances shall be computed.
-                    Note that this will produce a pairlist different from the pairlist produced by :py:func:`pairs` in that this does not exclude
-                    1-2 neighbors.
 
         indices2: iterable of integers (either list or ndarray(n, dtype=int)), optional:
                     Only has effect if :py:obj:`indices` is an iterable of integers. Instead of the above behaviour,
@@ -934,7 +932,7 @@ class MDFeaturizer(object):
 
     def add_distances_ca(self, periodic=True):
         """
-        Adds the distances between all Ca's (except for 1- and 2-neighbors) to the feature list.
+        Adds the distances between all Ca's to the feature list.
 
         """
         distance_indexes = self.pairs(self.select_Ca())
@@ -957,8 +955,6 @@ class MDFeaturizer(object):
 
                 iterable of integers (either list or ndarray(n, dtype=int)):
                     indices (not pairs of indices) of the atoms between which the inverse distances shall be computed.
-                    Note that this will produce a pairlist different from the pairlist produced by :py:func:`pairs`
-                    in that this does not exclude 1-2 neighbors.
 
         indices2: iterable of integers (either list or ndarray(n, dtype=int)), optional:
                     Only has effect if :py:obj:`indices` is an iterable of integers. Instead of the above behaviour,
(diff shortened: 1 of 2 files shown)


franknoe:
Aren't the tests just failing because they assume excluded_neighbors=2? So they should work if you just call pairs with excluded_neighbors=2 in the tests.

I'm all for doing it right immediately. The reason we had pairs in the beginning was to compute residue pairs, and then it makes sense to exclude one or two neighbors. But I think pairs is used in a more general way, e.g. for computing atoms pairs, and then excluding 1- and 2-neighbors might mean nothing, because they might not be neighbors in the structure.

gph82:
"Aren't the tests just failing because they assume excluded_neighbors=2? So they should work if you just call pairs with excluded_neighbors=2 in the tests."

It's not that I don't know how to make the tests work, sorry, i meant that we're essentially changing the api call. Now we don't have that many users, but I guess one should avoid it, right? I'll update the test and the default before we merge.

franknoe:
Yes, one should be careful with changes, but I think it's OK to change 
default behavior when it makes sense. That will not crash people's codes 
and make the data processing only minimally more expensive. In our 
initial implementation of the pairs function we simply haven't though 
about this.

Am 03/08/15 um 22:03 schrieb Guillermo Pérez-Hernández:

> "Aren't the tests just failing because they assume 
> excluded_neighbors=2? So they should work if you just call pairs with 
> excluded_neighbors=2 in the tests."
> 
> It's not that I don't know how to make the tests work, sorry, i meant 
> that we're essentially changing the api call. Now we don't have that 
> many users, but I guess one should avoid it, right? I'll update the 
> test and the default before we merge.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/479#issuecomment-127390474.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
on it

gph82:
we should at least put it in the changelog. Are we systematically doing changelog entries, or on a case-by-case basis?

franknoe:
sure, put it in. So far the changelog has been mostly ad hoc and 
described the general changes, but I think Martin has dug up a way to 
generate changelogs out of commit messages.

Am 03/08/15 um 22:16 schrieb Guillermo Pérez-Hernández:

> we should at least put it in the changelog. Are we systematically 
> doing changelog entries, or on a case-by-case basis?
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/479#issuecomment-127395111.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
The commit message was wrong. Now we can merge

gph82:
I am a bit confused as to why this PR now shows also @fabian-paul 's changes...

gph82:
Hi, can someone merge this? This keeps me from developing smaller fixes on my devel branch

franknoe:
I have to wait for the AppVeyor build. That can unfortunately take a while - apparently the freely available server resources are insufficient.

franknoe:
What about leaving the add_distances_ca() unchanged by adding excluded_neighbors=2 as a kwarg and passing that to the neighbor calculation? For Ca's it does make sense to leave out neighbors because they are ordered.

gph82:
At this moment, the complexity of doing this in my tree is much larger than merging and doing that request again. I don't have mergetools in my system (because of an update)

gph82:
thanks!

PR #494, opened 2015-08-14
[util/annotators] simplified and extended shortcut. Added new impl for class method aliases.

marscher:
- shortcut decorator may only be used for functions
- alias/aliased decorators are only for use with classes

Due to my research, there is no reliably working way (at least if one wants to support multiple python versions) to determine if a function is a class member, **before** one has an instance of this class. Since we want these aliases before instance creation, this is the only way it can work.

Above Python3.3 there is a way to obtain the class of an unbound method, but all prior versions lack the language features to obtain it (easily). Dirty hacks are not desired.

### DIFF
--- a/pyemma/msm/estimators/_dtraj_stats.py
+++ b/pyemma/msm/estimators/_dtraj_stats.py
@@ -3,11 +3,11 @@ __author__ = 'noe'
 import numpy as np
 
 from msmtools import estimation as msmest
-from pyemma.util.annotators import shortcut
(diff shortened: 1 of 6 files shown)


marscher:
Also it would be beneficial to warn the programmers using "shortcut" for methods, but this would involve solving the unbound method problem again. So please be careful, when you use this. The shortcut decorator was actually never working for class methods and cluttered the **all** variable of the classes module with aliases to the decorated method.

PR #497, opened 2015-08-20
Added continuous contact feature

ajkluber:
Added a continuous contact feature that emulates ContactFeature but calculates "contact" using a tanh function.


franknoe:
Thanks a lot @ajkluber for contributing! The code looks good, but I wanna think about a way to generalize these things in order to avoid having an explosion of the number API functions. Switching functions such as this one are related to the definition of basis sets.

Let's do that after the next release, so I'll keep the PR open for now.

marscher:
@ajkluber the tests are failing now. I figured out that the distance calculation with numpy and mdtraj yield different results, but there is also another error. Can you please have a look?

franknoe:
I still have the feeling that adding a function tanh contacts is too specific, so we should either generalize that to arbitrary functions (e.g. the user could give a pointer to a function to be applied on the distance), or leave it to be implemented via CustomFeatures.
Since there is a failure anyway, can we revert this? We can also discuss with Alex in two weeks what would be a good way of presenting this feature while keeping a general API.

marscher:
Agreed, I will revert it. Implementing this via a custom feature definition is also easy (and has the same performance).

franknoe:
Maybe we can add this as an example to the custom feature function.

Am 12/02/16 um 20:10 schrieb Martin K. Scherer:

> Agreed, I will revert it. Implementing this via a custom feature 
> definition is also easy (and has the same performance).
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/497#issuecomment-183453453.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

ajkluber:
Yeah I agree that is probably best.

PR #502, opened 2015-08-25
[coordinates]: wrapped docs and improved kmeans docs

franknoe:


franknoe:
@marscher : why does AppVeyor build cancel?

franknoe:
Apparently it went through now. Never mind.

PR #508, opened 2015-08-26
[progressbars] auto close finished progress bars after 2s. 

### LINKED ISSUE #501: Implied Timescales with error estimation generates many progressbars
Implied timescales with error estimation shows a progress bar for each timescale. This is not desired. I think the problem is that progress bars are created rather low-level, but the decision of now progress bars are shown should be made on the highest level (at the API function). 

In this case, can the ImpliedTimescales estimator switch off the individual progress bars for the estimator it uses and instead just show one progress bar which steps by lag times computed?

What would be cooler (but probably more complicated) is that the high-level function (ImpliedTimescales) allocates a progress bar, and the called functiones (BayesianMSM) just fill a section of that bar given to it.

marscher:
Func estimate_param_scan shows progress, Fixes #501

### DIFF
--- a/pyemma/_base/estimator.py
+++ b/pyemma/_base/estimator.py
@@ -1,19 +1,18 @@
 from __future__ import absolute_import
-from six.moves import range
 
-__author__ = 'noe'
+from six.moves import range
+import inspect
+import joblib
 
 from pyemma._ext.sklearn.base import BaseEstimator as _BaseEstimator
 from pyemma._ext.sklearn.parameter_search import ParameterGrid
 from pyemma.util.log import getLogger
 from pyemma.util import types as _types
 
-import inspect
-import joblib
-
-
 # imports for external usage
 from pyemma._ext.sklearn.base import clone as clone_estimator
+
+__author__ = 'noe'
 
 def get_estimator(estimator):
     """ Returns an estimator object given an estimator object or class
@@ -28,8 +27,10 @@ def get_estimator(estimator):
 
     """
     if inspect.isclass(estimator):
-        estimator = estimator()  # construct the estimator with default settings
+        # construct the estimator with default settings
+        estimator = estimator()
     return estimator
+
 
 def param_grid(pargrid):
     """ Generates an iterable over all possible parameter combinations from the grid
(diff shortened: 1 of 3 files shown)


franknoe:
Doesn't seem to change anything. I still get a separate progress bar for each lag estimation. So this does not fix #501.

screen shot 2015-08-27 at 17 20 42

marscher:
For implied timescales of SampledModels, one gets len(lags) \* n_samples 
estimations, which should go into bar - alright.

So if I set self.estimator.progress_silence = True if self.estimator is 
a SampledModel, the flag gets lost during clone() from base.sklearn. I 
do not understand why.

Do I have to set this flag via set_params somehow, which would be 
unhandy, since every progress reporting class can not simply rely on the 
logic of deriving ProgressReporter, but also has to make sure the 
derived attributes do not get lost on the way.

franknoe:
Because clone copies all Estimator parameters. But Estimator parameters are defined by the **init** of the Estimator class, i.e. for the Bayesian HMSM it's only those:

``` python
    def __init__(self, nstates=2, lag=1, stride='effective', prior='mixed', nsamples=100, init_hmsm=None,
                 reversible=True, connectivity='largest', observe_active=True, dt_traj='1 step', conf=0.95):
```

If you want to transport information about progress bars across cloning, you could add something like show_progress (or something else like progressbar ...) as a variable here. That must then be copied into a class member self.show_progress. I would not use progress_silence because that's by itself not a very informative name. Similar for BayesianMSM. 

This feels like a dirty solution to me, but I'd first have to understand the entire progress bar concept to propose a clean solution. At least the approach above should work.

marscher:
The approach is the following:
1. A progress reporting class A derives from ProgressReporter
2. A registers a progress with self._progress_register(batch_size, 
   description, stage)
3. A updates its progress via self._progress_update(increment, stage)

The ProgressReporter takes care of ProgressBar handling.

Why do you think it is a "dirty" solution to have show_progress as a 
parameter?

franknoe:
It feels dirty to make logging information an Estimation parameter. But doesn't matter, let's implement it like this for now.

Issue #487, opened 2015-08-05
[add_distances_ca] change default behaviour


gph82:
Request from @franknoe What about leaving the add_distances_ca() unchanged by adding excluded_neighbors=2 as a kwarg and passing that to the neighbor calculation? For Ca's it does make sense to leave out neighbors because they are ordered.

marscher:
fixed by #605

PR #514, opened 2015-08-28
Add distances ca exclude neighbors

gph82:
This addresses #487 . I did not go with the kwargs implementation, dunno if it can be kwarg and a default simultaneously. Also, I changed add_distances_ca so that the definition of neighbor is done with the residue indices and not the atom indices. This way, the starting geometry can be an all atom geometry.


gph82:
BTW, the tests need these extra files to check add_distances_Ca in the presence of other atoms, so I added them (each file has one frame). Perhaps we want a different way?

franknoe:
Can we use the same keyword as elsewhere (excluded_neighbors)? It's clear from the context that Ca's are meant. 

pdb's are not ideal because they are counted as code lines (our statistics are already really messed up, but doesn't mean we have to carry on with it). I suggest to zip them and then open them with python's zip package - makes them not count as code and also keeps them smaller (which is good for the repo). But perhaps @marscher has a better idea.

Otherwise fine. #487 is not crucial for the release, so this is not urgent.

gph82:
Here's the refactor. I'll wait for @marscher 's for the .zip thing. Not urgent

gph82:
mdtraj can load pdb.gz files, which made this PR easier. This can be merged now, I think

gph82:
residue_subset parameter added.

I know that we have feat.select or feat.add_distances for arbitrary selections. But the notion of "distance-mapping" particular regions (=residue_subset) is more in line with use cases, where the "all or nothing" behavior of add_distances_Ca is a bit limiting.

gph82:
Any feedback on this, particularly this new parameter (https://github.com/gph82/PyEMMA/commit/5a20cb4b80c92f4fd342f6df0b4406dce38f41e4) are welcome

gph82:
Hi, if we're not sure about the new parameter, at least we should merge the correct handling of excluded neighbors, i think

franknoe:
I think this one has been forgotten. Is "residue_subset" meant such that you will only consider distances between Ca's within this set? Or Ca's within this set with all Ca's?
Moreover, is this extra parameter only in the Ca-Distance or in all distance computations? 

Note that Ca-distance function was just meant as a convenience shortcut to the ordinary add_distance function with Ca's. You could have the same effect by using add_distance with a selection containing only the Ca's. That means conceptually, parameters simplifying distance calculations should go into add_distance, and then add_distance_Ca should just call add_distance with the Ca-selection and copy all relevant parameters in the add_distance call.

franknoe:
plus: the exclude_neigbhor merge needs to wait til after the 2.0.2 release because it slightly changes the API behavior. So that would go into 2.1

gph82:
"I think this one has been forgotten. Is "residue_subset" meant such that you will only consider distances between Ca's within this set? Or Ca's within this set with all Ca's?"
The first one:

``` python
#        residue_subset : iterable of integers, default is None
#           Only the distances between atoms of the given residues will be added to the feature list.
#          This parameter is 1-indexed, coinciding with the resSeq record of the mdtraj topology
#         (http://mdtraj.org/latest/atom_selection.html#keywords)
```

So far, it's only here, but can be extended to all things distance, as you say.

What's more important to me is that the excluded_neighbors parameter is using atom indices (and not their residue indices), meaning that it's meaningless in geometries with more atom types (which are precisely the ones for which add_distances_Ca makes sense). This is what this commit is doing:
https://github.com/gph82/PyEMMA/commit/732e1f10808a91cbc033b4fc2ebafcb7b927d0ca.

franknoe:
I'm confused. Is that commit part of this PR or has it already been merged?

gph82:
It's part of this pr

franknoe:
Just to make sure I understand you correctly: Before this PR (i.e. in the current devel), both `add_distances()` and  `add_distances_ca()` had a parameter `exclude_neighbors` with settings 0 and 2, respectively. Are you saying that in `add_distances_ca()` this parameter is used the wrong way, because it has just been passed on to `add_distances()`, such that the exclusion does not affect other Ca's, but rather other atoms that are next to Ca's. And your PR corrects that such that now `exclude_neighbors` in `add_distances_ca()` really does map to the next Ca's?

Correct?
If that is so, that part can be merged. I also just saw that exclude_neighbors used to be 2 before, so it doesn't affect the API but rather just correct a bug.

Can we separate this from the subset parameter. I think that one would be better addressed in a general way.

gph82:
yeap, that's exactly it. I'll create a new PR only with that part

franknoe:
ok, thanks!

Am 01/11/15 um 14:10 schrieb Guillermo Pérez-Hernández:

> yeap, that's exactly it. I'll create a new PR only with that part
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/514#issuecomment-152824625.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
All commits except https://github.com/gph82/PyEMMA/commit/5a20cb4b80c92f4fd342f6df0b4406dce38f41e4 are now in #605 , so I'm closing this

PR #517, opened 2015-08-28
pin down bhmm to version 0.5.1

### LINKED ISSUE #516: require exact bhmm version for PyEMMA 2.0
The bhmm API is likely to change. Thus PyEMMA 2.0 needs to get a hard requirement to the exact current bhmm release in order to be functional in the future. We can relax that constraint once the bhmm API is stable.

marscher:
Fixes #516

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -238,7 +238,7 @@ metadata = dict(
                       'mdtraj>=1.4.0',
                       'matplotlib',
                       'msmtools',
-                      'bhmm',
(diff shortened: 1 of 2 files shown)


franknoe:
Thanks!

franknoe:
Should it be == instead of = ?

marscher:
in setup.py its ==, in conda recipes it is = ...

marscher:
Actually you're right :-)

PR #523, opened 2015-08-30
[config] allow setting by attribute and add docstring examples.

marscher:


franknoe:
That's OK.

franknoe:
Now it works, thanks.

PR #524, opened 2015-08-30
[LogHandling] Garbage collect unused Logger instances.

marscher:
Uses weakrefs on each Transformer instance with a callback function to remove
the corresponding logger. Otherwise these loggers accumulate and never gets removed for one session, which may be cause problems if one creates tons of Estimators/Transformers.


marscher:
This also unifies the way loggers are created for Transformers, Estimators and Featurizer.

PR #526, opened 2015-08-30
rewrote logging settings handling

marscher:
- a fully flexible logging.yml (YAML) has been added to pyemma dir (default settings: loglevel INFO, stdout, no file logging)
- a copy of logging.yml is created in users .pyemma dir and used in favor of default file
- replaced [Logging] section from pyemma.cfg with logging_config directive (file or DEFAULT keyword)

This can be safely postponed to 2.1


franknoe:
This is no longer automatically mergeable, so better if @marscher has another look on it.

marscher:
rebased the branch on latest devel

Issue #520, opened 2015-08-28
k-means warning


franknoe:
With the new memory behavior of using a memmap when the RAM is insufficient, do we will need this message?

```
2015-08-29 01:25:49,683 coordinates.clustering.KmeansClustering[61] WARNING  K-means implementation is currently memory inefficient. This calculation needs 132 megabytes of main memory. If you get a memory error, try using a larger stride.
```

clonker:
I guess the warning is at the wrong place (and with the wrong text).. maybe one should warn when in-memory mode gets switched to memmapped mode.

marscher:
yes, that makes more sense, because that will be comparably slow.

Issue #470, opened 2015-07-31
Set reasonable default for number of clusters


franknoe:
I suggest to set sqrt(N) as a default for the number of clusters in clustering algorithms. N is the total number of samples. If anyone feels that we should limit the number of clusters we cut that off to max 5000 (that would be reached for 25 M frames, which is quite a bit).

Should N be counted with or without stride? Good question, in principle I would say N is the number of assigned frames, but of course we have to make the decision before assigning. So I guess I would take the stride into account.

PR #543, opened 2015-09-02
Kmeans memory requirement management and default number of centers

clonker:
This PR deals with issues #470 and #520. 
- Removed the warning if kmeans needs more than 100MB of memory, adding a warning if the memory requires exceed what is available. In order to have the information available, psutil was added to the required packages list. 
- Changed the api default number of cluster centers to 'None', which results in a number of cluster centers chosen by min(sqrt(N), 5000), where N is the number of considered frames, i.e., takes stride into account.

### DIFF
--- a/pyemma/coordinates/api.py
+++ b/pyemma/coordinates/api.py
@@ -1206,7 +1206,8 @@ def cluster_kmeans(data=None, k=None, max_iter=10, tolerance=1e-5, stride=1,
         input data, if available in memory
 
     k: int
-        the number of cluster centers
+        the number of cluster centers. When not specified (None), min(sqrt(N), 5000) is chosen as default value,
+        where N denotes the number of data points
 
     max_iter : int
         maximum number of iterations before stopping. When not specified (None), min(sqrt(N),5000) is chosen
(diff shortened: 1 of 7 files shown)


marscher:
Thank you very much! Looks good to me.

franknoe:
Is the number of cluster centers specific to k-means or does it apply to all clustering algorithms? I think it could be applied to all clustering algorithms, and the default could be implemented in the generic clustering interface if that's code-wise possible.

clonker:
One could add another "abstract-class-layer" for clustering algorithms that take the number of cluster centers as input parameter. For other clustering algos (like regspace), the number is unknown a prior. Should I do that?

franknoe:
Ah, you're right. No, that's too little code to justify another class 
layer. Let's just implement the same default for uniform time clustering.

Thank you very much.

Am 02/09/15 um 16:35 schrieb Moritz Hoffmann:

> One could add another "abstract-class-layer" for clustering algorithms 
> that take the number of cluster centers as input parameter. For other 
> clustering algos (like regspace), the number is unknown a prior. 
> Should I do that?
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/543#issuecomment-137104987.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

Issue #532, opened 2015-08-30
Unify Estimator concept and better sklearn compatibility


franknoe:
Unify Estimator concept across PyEMMA. This goes along with a better compatibility with sklearn. Here are some raw ideas that need to be discussed and re-evaluated:
- Replace (using deprecation) `Transformer.parametrize()` by `Transformer.estimate(X, kwargs)`. I'm not sure if this will have any undesired side-effects on our pipeline handling because in this concept setting the data producer and running the estimation occur in the same step, while we currently do it subsequently.
- Our present `Transformer` class is really an Estimator and a Transformer at the same time (these concepts are also not clearly distinguished in sklearn). So our object presently called `Transformer` might derive from `Estimator` (PyEMMA base class). It is different from the estimators in the msm package by producing a Transformer instead of a kinetic model (which might not be a problem, see next point), and by accepting trajectories and trajectory data sources as input and streaming/chunking them instead of assuming that data lives in memory (as for e.g. `MaximumLikelihoodMSM`). Perhaps this is really a subclass of `Estimator` such as `TransformEstimator`.
- Suggest to have a separate class `Transformer` which is not necessarily estimated (e.g. a `LinearTransform` or `AffineTransform` could be just defined by the user, or could be estimated from PCA or TICA). Suggest that `Transformer` is a subclass of `Model`, such that the simple concept "An Estimator produces a Model" always applies, and we can use the benefits of having a generic Model class (which really is any sort of static thing with parameters) as in Transformers.
- `Estimator` is already largely consistent with sklearn. Still missing: we currently have no explicit y argument in fit(X).
- `TransformEstimator` will derive from sklearn `TransformerMixin` and offer `fit_transform(X, kwargs)`

marscher:
In my opinion we have fulfilled most of this. Support for partial_fit can be added later.

franknoe:
yes

PR #545, opened 2015-09-02
Decorator to check if estimator is estimated

clonker:
This PR contains a decorator that can be applied to estimator methods that need an estimated state to return meaningful results, see issue #532. Also it contains a modification of the deprecated annotation allowing to deprecate class methods as well (parenthesis required then, see included test case).
Last, the "timescales" property-method of "implied_timescales" has been decorated with "estimation_required".

### DIFF
--- a/pyemma/util/annotators.py
+++ b/pyemma/util/annotators.py
@@ -37,6 +37,8 @@ from __future__ import absolute_import
 from functools import wraps
 import warnings
 from six import PY2
+from decorator import decorator
+from pyemma._base.estimator import Estimator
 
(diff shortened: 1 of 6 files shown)


pyemma/util/annotators.py
```diff
@@ -37,6 +37,8 @@ def foo(self):
 from functools import wraps
 import warnings
 from six import PY2
+from decorator import decorator
+from pyemma._base.estimator import Estimator
```
marscher, line 41: Estimator is never used in this module, right?

PR #553, opened 2015-09-03
Mdtraj topology

jhprinz:
Allow using an mdtraj.Topology object when initializing a featurizer.

For convenience and integration of PyEmma into OpenPathSampling.

### DIFF
--- a/pyemma/coordinates/tests/test_featurizer.py
+++ b/pyemma/coordinates/tests/test_featurizer.py
@@ -245,7 +245,7 @@ class TestFeaturizer(unittest.TestCase):
         self.assertEqual(len(desc), self.feat.dimension())
 
     def test_backbone_dihedrals(self):
-        self.feat = MDFeaturizer(topfile=self.asn_leu_pdbfile)
+        self.feat = MDFeaturizer(topology=self.asn_leu_pdbfile)
         self.feat.add_backbone_torsions()
 
         traj = mdtraj.load(self.asn_leu_pdbfile)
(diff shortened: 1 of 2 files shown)


marscher:
I like the idea of directly passing topolgy objects, but I vote against breaking the api for that. Of course you're right that the name is not correct anymore, but this change will break everybody's code.

marscher:
you should also add a single test this case

franknoe:
why does this break code? This is not a kwarg.

marscher:
all parameters in Python can also be passed by their name, if you do so and the parameter name changes, the code breaks.

pyemma/coordinates/tests/test_featurizer.py
```diff
@@ -245,7 +245,7 @@ def test_dihedrials_cossin(self):
         self.assertEqual(len(desc), self.feat.dimension())
 
     def test_backbone_dihedrals(self):
-        self.feat = MDFeaturizer(topfile=self.asn_leu_pdbfile)
+        self.feat = MDFeaturizer(topology=self.asn_leu_pdbfile)
```
marscher: example for API breakage, because if arg is passed by the old name, this will fail

franknoe:
Indeed! I didn't know. But why would one do that, i.e. call a mandatory 
parameter by kwarg?

Am 03/09/15 um 16:23 schrieb Martin K. Scherer:

> all parameters in Python can also be passed by their name, if you do 
> so and the parameter name changes, the code breaks.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/553#issuecomment-137464919.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
maybe to be able to pass in further arguments later on in arbitrary order? There may be many reasons, but actually it does not matter, we just need to preserve it.

PR #556, opened 2015-09-10
[coor.api.tica] parameter epsilon is included in the api call

gph82:
I know we're reluctant to changes in the API, but I think we need this. 

If we're uncomfortable, the softer version of this would be simply to include the SVD behaviour in the API docu, so that users know why they might be getting less TICs than could, in principle, be produced.


marscher:
- First of all: you have deleted the ipython submodule. Please fix that.
- Put the parameter at last in order not break the already existing order.

franknoe:
I've just discussed this with Guille. The point is that we can never guarantee that all dimensions will be contained in the output when we want a numerically stable solution, because the GEV-Problem is practically always rank-deficient. Instead of pulling this parameter to the API level, we should document this behavior and perhaps include a discussion about it in a TICA notebook.

Closing this PR...

PR #573, opened 2015-10-06
[plot_implied_timescales] different units and dt for x and y axis

gph82:
This allows for generating the ITS plot using different labels (parameter 'units') and different timesteps (parameter 'dt') for the x and y axis. One (or both) of these parameters can be a list of two elements, mapping to the x,y-axis. The use-case is something like the attached picture, where using the same unit in both axes is not that clear. 

snap4


franknoe:
thanks

Issue #565, opened 2015-09-25
Distance units


stefdoerr:
Are the units produced by featurizer.add_distance actually in nanometers? I assumed it would return distances in Angstrom but it consistently reports me 1/10th of the Angstrom distance I get in my code.
Is it maybe an issue with reading the xtc files?

stefdoerr:
Assuming the featurizer returns distances in the units used in the trajectory file, you have a bug in your xtc reader. 
VMD shows the distances correctly/consistently with my code.

franknoe:
I think this is an issue to be answered by Guillermo. We use mdtraj for 
I/O and most of the featurization, which I think uses nm by default, and 
we just pass those distances on.

Am 25/09/15 um 14:58 schrieb Stefan:

> Are the units produced by featurizer.add_distance actually in 
> nanometers? I assumed it would return distances in Angstrom but it 
> consistently reports me 1/10th of the Angstrom distance I get in my code.
> Is it maybe an issue with reading the xtc files?
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/markovmodel/PyEMMA/issues/565.

## 

---

Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

## Mail:  Arnimallee 6, 14195 Berlin, Germany

gph82:
Hi, it's Nanometers, the units of mdtraj

On 09/25/2015 03:43 PM, Frank Noe wrote:

> I think this is an issue to be answered by Guillermo. We use mdtraj for
> I/O and most of the featurization, which I think uses nm by default, and
> we just pass those distances on.
> 
> Am 25/09/15 um 14:58 schrieb Stefan:
> 
> > Are the units produced by featurizer.add_distance actually in
> > nanometers? I assumed it would return distances in Angstrom but it
> > consistently reports me 1/10th of the Angstrom distance I get in my code.
> > Is it maybe an issue with reading the xtc files?
> > 
> > —
> > Reply to this email directly or view it on GitHub
> > https://github.com/markovmodel/PyEMMA/issues/565.
> 
> ## 
> 
> ---
> 
> Prof. Dr. Frank Noe
> Head of Computational Molecular Biology group
> Freie Universitaet Berlin
> 
> Phone: (+49) (0)30 838 75354
> Web: research.franknoe.de
> 
[...]

PR #574, opened 2015-10-08
[add_contacts] threshold in nm instead of AA

gph82:
Undecided on this one:

When speaking about contacts, people usually use Angstrom. However, mdtraj deals and produces distances and positions in nm (and deals with different formats internally). So I propose we stick to nms also when inputting parameters. 

This addresses  #565


gph82:
I'd like to merge this minor change if it's okay

PR #577, opened 2015-10-10
[plots.networks] **textkwargs for pretty printing of labels

gph82:
Even more beatifying through a **textkwargs for the annotation object. 
tpt1


gph82:
The above call looks:

``` python
        __, pos = pyemma.plots.plot_flux(iTPT, 
                                         pos=ipos,                                
                                         state_labels=[1,'open','3CAP','D\n(75%)'], 
                                         state_colors=state_colors, show_frame=True,
                                         color='pink', multialignment='right', rotation=45
                                        )     
```

PR #591, opened 2015-10-27
[coor/traj_info_cache] added support for .npy

marscher:
Assuming you're opening the same .npy files, their lengths (first dim) is being
cached in the same manner as trajectory files.


franknoe:
OK for me.

PR #592, opened 2015-10-27
fixes #585

### LINKED ISSUE #585: featurizer._pair_wise_input uses itertools.combinations
This function is intended to iterate over the output (generator pattern), but to create an index array it is prohibitively memory consuming, since the intermediate has to be converted to a Python list (which will create a Python object for every index involved in the pairing).

There must be a simple pure numpy solution to this.

marscher:


franknoe:
I am worried about this line:

```
+    from sklearn.utils.extmath import cartesian as _product
```

We don't want full sklearn dependence. We have only included their base module.

marscher:
right, I just fixed this.

PR #594, opened 2015-10-27
added signal handling to enable easy debugging of hung python processes.

marscher:
Scenario: a long-running cell evaluation in a notebook or plain python process has to be investigated (non-responsive, no useful console/logging).

You can either attach a debugger or at least at which state the program is at the moment.

The newly introduced debug module in util will enable this for you:

``` python
import os
print os.getpid() # prints eg. 2000
from pyemma.util import debug
debug.register_signal_handlers()

while True:
    import time
    time.sleep(1)
```

Now you can send the signal to log a stack trace to file/stdout.

``` bash
kill -42 2000
```

or attach the debugger to the frame which is on top of the stack, when the signal is received and handled by Python.

``` bash
kill -43 2000
```

I've chosen these signal numbers, because IPython already reserved SIGUSR{1,2} and it is unlikely that other packages using these.

### DIFF
--- a/pyemma/util/debug.py
+++ b/pyemma/util/debug.py
@@ -1,0 +1,48 @@
+''' registers some debug handlers on the importing Python process:
+1. SIGUSR1 -> show a stack trace of the current frame
+2. SIGUSR2 -> attach the Python debugger pdb to the current process
+
+
+
+Created on 15.10.2015
+
+@author: marscher
+'''
+import signal
+from log import getLogger
+
+logger = None
+
+
+def show_stacktrace(sig, frame):
+    import traceback
+    import StringIO
+
+    global logger
+    if logger is None:
+        logger = getLogger('dbg')
+
+    out = StringIO.StringIO()
+
+    traceback.print_stack(frame, file=out)
+
+    out.seek(0)
+    trace = out.read()
+    logger.error(trace)
+
+
+def handle_pdb(sig, frame):
+    import pdb
+
+    pdb.Pdb().set_trace(frame)
+
+
+def register_signal_handlers():
+    # Register handler
+    global logger
+
+    signal.signal(42, show_stacktrace)
+    signal.signal(43, handle_pdb)
+
+def unregister_signal_handlers():
+    pass
--- a/pyemma/util/debug.py
+++ b/pyemma/util/debug.py
@@ -8,8 +8,10 @@ Created on 15.10.2015
 
 @author: marscher
 '''
+from __future__ import absolute_import
+
 import signal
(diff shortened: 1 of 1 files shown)


franknoe:
I don't fully understand this, so I suggest somebody who does reviews it.

In any case, please fully document all API methods (there are no docstrings here).

marscher:
Added documentation as requested. IMO this is merge-able now.

franknoe:
Yes, thanks! AppVeyor still failing, so I guess we need to fix the source of that failure first.

marscher:
Its failing because of this (unrelated) issue:

ImportError: DLL load failed: The specified module could not be found.

franknoe:
So what is this failure due to? I've never seen it.

marscher:
Me neither. The warnings shown in the output suggests that it is related to the QT installation procedure (which is a dependency of matplotlib).

franknoe:
So does our devel branch also fail because of this? Either way we need to fix it.

marscher:
devel fails because of the different results in hmm #600 (bhmm 0.5.2)

franknoe:
that's expected (see last comment in #600), but this means there is 
either something else different between this fork and devel or that this 
issue is a random failure of appveyor. Start appveyor again after fixing 
#600?

Am 30/10/15 um 18:06 schrieb Martin K. Scherer:

> devel fails because of the different results in hmm #600 
> https://github.com/markovmodel/PyEMMA/pull/600 (bhmm 0.5.2)
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/594#issuecomment-152588388.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
I think it is a bug in the matplotlib package which is only triggered in conda-build. So we would have to wait until it is fixed upstream.

franknoe:
If it's a matplotlib failure, devel should exhibit the same problem.

If it's due to a new matplotlib release, we could exclude that 
matplotlib version. I guess this problem only occurs on windows?

Am 30/10/15 um 18:12 schrieb Martin K. Scherer:

> I think it is a bug in the matplotlib package which is only triggered 
> in conda-build. So we would have to wait until it is fixed upstream.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/594#issuecomment-152589738.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

PR #598, opened 2015-10-30
changes added to the changelog

gph82:
I went through my commits and this is the summary. The new features (minrmsd, groupmindist, resmindist) were already in the logs of their respective versions versions (1.2.2 and 2.0).

If it's too verbose feel free to crop at your discretion.


marscher:
Thank you! I think the level of verbosity is just adequate for end-users :+1:

PR #608, opened 2015-11-04
added support for bool trajectories

stefdoerr:
modified types.py to accept bool trajectories so that I can feed TICA a list of np.ndarrays with dtype=bool

### DIFF
--- a/pyemma/util/types.py
+++ b/pyemma/util/types.py
@@ -119,6 +119,15 @@ def is_float_matrix(l):
             return True
     return False
 
+def is_bool_matrix(l):
+    r"""Checks if l is a 2D numpy array of bools
+
+    """
+    if isinstance(l, np.ndarray):
+        if l.ndim == 2 and (l.dtype == bool):
+            return True
+    return False
+
 def is_float_array(l):
     r"""Checks if l is a numpy array of floats (any dimension
 
@@ -444,7 +453,7 @@ def ensure_traj(traj):
     r"""Makes sure that dtraj is a discrete trajectory (array of float)
 
     """
-    if is_float_matrix(traj):
+    if is_float_matrix(traj) or is_bool_matrix(traj):
         return traj
     elif is_float_vector(traj):
         return traj[:,None]


marscher:
Looks fairly simple. The input is converted to float32 implicitly in the spots where it is needed.

@franknoe If I'd add a little test for this, would you accept it for v2.0.2?

franknoe:
I guess the PR is OK and doesn't break anything. I understand it allows you to smuggle booleans into TICA.

What I'm wondering is what happens with this input inside TICA. If the TICA transformer gets X also as boolean array, we would run into the problem that boolean multiplications are very slow (that was discussed before). If X is a float at this point, it must be that the booleans are being transformed by the data-in-memory reader. Martin, can you comment on this?
@stefdoerr , have you tested the speed performance of your code?

marscher:
Of course we can enforce the conversion in NumPyFileReader and DataInMemory. But I'd guess that Numpy already converts arguments to float in mixed datatype operations.

marscher:
``` python
X=np.random.randint(0,100, size=(10, 10000))
Y=np.random.random((10000,10))
Xf = X.astype(float)
assert Xf.base is None

%timeit  Y.dot(X)
1 loops, best of 3: 927 ms per loop

%timeit  Y.dot(X.astype(float))
1 loops, best of 3: 929 ms per loop

%timeit  Y.dot(Xf)
1 loops, best of 3: 928 ms per loop
```

franknoe:
Yes, but we have something like C += X.dot(X.T), and that would be super-slow (try it with a smaller matrix unless you wanna crash your notebook). So it does matter in which form X arrives.

BTW, in the above you compute outer products (you get a 10000x10000 matrix) - in TICA we would rather do the transposed.

marscher:
You're right. So I'm going to add this conversion in the readers.

franknoe:
hm, that's a deeper intervention than Stefan's PR. Let's see what Stefan 
replies, I guess he has tested this. If he doesn't run into this speed 
bottleneck, I'd like to understand why.

Am 04/11/15 um 19:15 schrieb Martin K. Scherer:

> You're right. So I'm going to add this conversion in the readers.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/608#issuecomment-153815396.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

stefdoerr:
I looked at the TICA code and you are very careful to actually cast everything to float64 during calculations, so there is no problem with that. The only place you directly use the boolean matrix is to calculate sums and to subtract from it the mean (lines 271, 275, 278, 282, 302 of tica.py). Ran some tests and bool operations seem double as slow but still quite fast. Also bool - float obviously is cast to float.

``` python
a = np.zeros((1000,100), dtype=bool)
a2 = np.zeros((1000,100))
b = np.random.rand(100, 1000)

In [33]: %timeit np.sum(a)
10000 loops, best of 3: 104 µs per loop

In [34]: %timeit np.sum(a2)
The slowest run took 323.75 times longer than the fastest. This could mean that an intermediate result is being cached 
10000 loops, best of 3: 42.7 µs per loop

In [10]: %timeit -n10000 np.transpose(a) - b
10000 loops, best of 3: 202 µs per loop

In [11]: %timeit -n10000 np.transpose(a2) - b
10000 loops, best of 3: 129 µs per loop

In [26]: z = np.transpose(a) - b

In [29]: z.dtype
Out[29]: dtype('float64')
```

This minor slowdown could obviously be avoided if you just add X = X.astype(float) on line 245. Then all the rest will be float-float operations.

**HOWEVER**:
I tested it on TICA and it takes roughly the same time with float32 and bool. 8 minutes. 

But this brings me back to the other issue which was that 
a) even when I switched from float16 to float32 I didn't see any speedup in TICA (since you internally cast to float64 anyways)
b) 8 minutes is quite slow for such a small dataset.
But we should leave this for another issue discussion (might open an issue)

stefdoerr:
To summarize, total execution time stays the same and you internally cast nearly everything to float64 in TICA. I don't see a reason not to add the change as is, unless you want to also add to `tica.py _param_add_data()` a line at the beginning that converts X to float to avoid possible slowdowns which I cannot detect though in a real case example.

franknoe:
Thanks Stefan! Yes, the speed issue is another issue and we should 
definitely work on that. If there is not slowdown, I am fine with 
merging. Please check if PCA has the same behavior though.

Am 05/11/15 um 10:28 schrieb Stefan:

> I looked at the TICA code and you are very careful to actually cast 
> everything to float64 during calculations, so there is no problem with 
> that. The only place you directly use the boolean matrix is to 
> calculate sums and to subtract from it the mean (lines 271, 275, 278, 
> 282, 302 of tica.py). Ran some tests and bool operations seem double 
> as slow but still quite fast. Also bool - float obviously is cast to 
> float.
> 
> a=  np.zeros((1000,100),dtype=bool)
> a2=  np.zeros((1000,100))
> b=  np.random.rand(100,1000)
> 
> In [33]:%timeit np.sum(a)
> 10000  loops, best of3:104  µs per loop
> 
> In [34]:%timeit np.sum(a2)
> The slowest run took323.75  times longer than the fastest. This could mean that an intermediate resultis  being cached
> 10000  loops, best of3:42.7  µs per loop
> 
> In [10]:%timeit-n10000 np.transpose(a)-  b
> 10000  loops, best of3:202  µs per loop
> 
> In [11]:%timeit-n10000 np.transpose(a2)-  b
> 10000  loops, best of3:129  µs per loop
> 
> In [26]: z=  np.transpose(a)-  b
> 
> In [29]: z.dtype
> Out[29]: dtype('float64')
> 
> This slowdown could obviously be avoided if you just add X = 
> X.astype(float) on line 245. Then all the rest will be float-float 
> operations.
> 
> _HOWEVER_:
> I tested it on TICA and it takes roughly the same time with float32 
> and bool. 8 minutes.
> 
> But this brings me back to the other issue which was that
> a) even when I switched from float16 to float32 I didn't see any 
> speedup in TICA (since you internally cast to float64 anyways)
> b) 8 minutes is quite slow for such a small dataset.
> But we should leave this for another issue discussion (might open an 
> issue)
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/608#issuecomment-154002865.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

franknoe:
In addition, we should discuss if we could add a few performance unit 
tests that are automatically run for selected tests problems. The code 
has become sufficiently big that we do not only need to take care to 
improve performance, but also to make sure that we don't accidentially 
reduce performance somewhere through side-effects.

Am 05/11/15 um 10:53 schrieb Stefan:

> To summarize, total execution time stays the same and you internally 
> cast nearly everything to float64 in TICA. I don't see a reason not to 
> add the change as is, unless you want to also add to |tica.py 
> _param_add_data()| a line at the beginning that converts X to float to 
> avoid possible slowdowns which I cannot detect though in a real case 
> example.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/608#issuecomment-154010244.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

stefdoerr:
PCA works equally fast on bool and float32. 3 minutes for both. Results are exact same.

franknoe:
OK, we can go ahead then. I think we just have to test the windows 
version manually (the current windows matplotlib conda is broken, which 
is a different issue) and are then ready to release 2.0.2

Am 05/11/15 um 11:38 schrieb Stefan:

> PCA works equally fast on bool and float32. 3 minutes for both. 
> Results are exact same.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/608#issuecomment-154024573.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
@stefdoerr 
I'd guess your implementation of TICA is also a 2-pass algorithm (1. pass: calculate mean, 2 calculate covariances)? If you hold all data in memory, the Python overhead might take up till 15% of the runtime. You might consider increasing the chunksize prior parametrization to lower that value. I'd also be interested in comparing both codes to see if there is some opportunity to increase performance on our side.

We only use double precision to sum up the mean, covariances are single precision.

@franknoe: Having a reliable performance regression testing would be a nice thing to have. We can do this in our own Jenkins instance. I will test the windows version manually and then push the release.

franknoe:
Cool. Is coveralls.io broken? I've seen multiple times that it doesn't start.

marscher:
Indeed. I think it broke down, when we switched building via the conda-recipe. The coverage plugin expects the code (py/c files) in a certain place after execution, which does not exist any more at runtime of the plugin.
I open a separate issue for that - thanks.

franknoe:
@stefdoerr : in order to grasp the essence of a new issue: given you 
have your feature data in arrays, let's see where possible bottlenecks 
are - in particular if anything is slowed down by our data streaming 
logic. How does the performance of our TICA algorithm compare to a 
direct application of the linear algebra routines? I.e. something like.

``` python
Xm = X.mean
C0 = Xm[:-tau].T.dot(Xm:-tau) + Xm[tau:].T.dot(Xm[tau:]
Ctau = Xm[tau:].T.dot(Xm[:-tau])
Ctau += Ctau.T
```

The above code isn't tested, so consider it to be pseudocode. Also, it 
doesn't include the diagonalization, but I guess that is not the 
rate-limiting step - correct me if I'm wrong.

Am 05/11/15 um 10:28 schrieb Stefan:

> I looked at the TICA code and you are very careful to actually cast 
> everything to float64 during calculations, so there is no problem with 
> that. The only place you directly use the boolean matrix is to 
> calculate sums and to subtract from it the mean (lines 271, 275, 278, 
> 282, 302 of tica.py). Ran some tests and bool operations seem double 
> as slow but still quite fast. Also bool - float obviously is cast to 
> float.
> 
> a=  np.zeros((1000,100),dtype=bool)
> a2=  np.zeros((1000,100))
> b=  np.random.rand(100,1000)
> 
> In [33]:%timeit np.sum(a)
> 10000  loops, best of3:104  µs per loop
> 
> In [34]:%timeit np.sum(a2)
> The slowest run took323.75  times longer than the fastest. This could mean that an intermediate resultis  being cached
> 10000  loops, best of3:42.7  µs per loop
> 
> In [10]:%timeit-n10000 np.transpose(a)-  b
> 10000  loops, best of3:202  µs per loop
> 
> In [11]:%timeit-n10000 np.transpose(a2)-  b
> 10000  loops, best of3:129  µs per loop
> 
> In [26]: z=  np.transpose(a)-  b
> 
> In [29]: z.dtype
> Out[29]: dtype('float64')
> 
> This slowdown could obviously be avoided if you just add X = 
> X.astype(float) on line 245. Then all the rest will be float-float 
> operations.
> 
> _HOWEVER_:
> I tested it on TICA and it takes roughly the same time with float32 
> and bool. 8 minutes.
> 
> But this brings me back to the other issue which was that
> a) even when I switched from float16 to float32 I didn't see any 
> speedup in TICA (since you internally cast to float64 anyways)
> b) 8 minutes is quite slow for such a small dataset.
> But we should leave this for another issue discussion (might open an 
> issue)
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/608#issuecomment-154002865.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

Issue #611, opened 2015-11-05
mini_batch_kmeans does not converge in param_init on huge data sets


marscher:
Having a large data set (eg. 377046 frames converted to contact maps, which have a dimension of 6420 each), mini_batch_kmeans fails to converge in param_init (kmeans++ initialisation).

The memory usage wents up like this:
![mem_usage](https://cloud.githubusercontent.com/assets/170287/10970678/9735fc3c-83d0-11e5-8e44-8ac346ad4678.png)

The stacktrace (generated by the new debug tool) shows that it is stuck in param_init (draw_sample):

``` python
05-11-15 15:07:47 dbg          ERROR      File "/home/mi/marscher/miniconda/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/home/mi/marscher/miniconda/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/ipykernel/__main__.py", line 3, in <module>
    app.launch_new_instance()
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/traitlets/config/application.py", line 592, in launch_instance
    app.start()
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/ipykernel/kernelapp.py", line 403, in start
    ioloop.IOLoop.instance().start()
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/zmq/eventloop/ioloop.py", line 151, in start
    super(ZMQIOLoop, self).start()
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/tornado/ioloop.py", line 866, in start
    handler_func(fd_obj, events)
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/tornado/stack_context.py", line 275, in null_wrapper
    return fn(*args, **kwargs)
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/zmq/eventloop/zmqstream.py", line 433, in _handle_events
    self._handle_recv()
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/zmq/eventloop/zmqstream.py", line 465, in _handle_recv
    self._run_callback(callback, msg)
  File "/home/mi/marscher/miniconda/lib/python2.7/site-packages/zmq/eventloop/zmqstream.py", line 407, in _run_callback
[...]

PR #612, opened 2015-11-05
[coor/clustering/minibatch-kmeans] fix memory usage of draw_sample.

marscher:
Fix for #611


clonker:
Thank you, this does look better than the old code; hopefully eliminates the memory/performance issues!

marscher:
I'm pretty sure it will fix it, because it does not create tons of Python objects during this step. Will test this on my big dataset now.

PR #619, opened 2015-11-09
[setup|conda-recipe] fix dependencies

marscher:


marscher:
Everything is fine actually, but Py34 win-32 still suffers from matplotlib import bug. Going to merge this.

PR #624, opened 2015-11-10
[coor/data/featurizer]  string comparison for group inputs

gph82:
… when parsing pairwise inputs. Otherwise a warning regarding "element-wise comparison" will be produced.


franknoe:
Wait. == is equality, is is identity. But you want to compare equality. So I think it's wrong to change to 'is'.

See here:
http://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce

marscher:
I think the elementwise comparision warning is triggered by numpy. So you have to distinguish by input type of 'scheme'.

gph82:
I think this is the correct way of doing it

PR #625, opened 2015-11-10
[coor/feature_reader+cache] workaround for missing __len__ function for xyz files.

### LINKED ISSUE #621: reading in of .xyz trajectory fails
I have a problem with reading in .xyz trajectory files. 
The example I am trying to run looks like:

```
import pyemma
import pyemma.coordinates as coor
import pyemma.msm as msm
import pyemma.plots as mplt
import numpy as np

print pyemma.__version__

topfile = 'adp.pdb'
traj_list = ['chain.xyz']
feat = coor.featurizer(topfile)
feat.add_dihedrals([[14,16,0,2],[12,14,16,0]],deg=True,cossin=False)
inp = coor.source(traj_list, feat)
print 'number of trajectories = ',inp.number_of_trajectories()
print 'trajectory length = ',inp.trajectory_length(0)
print 'trajectory time step = ',500.0 / (inp.trajectory_length(0)-1),'ns'
print 'number of dimension = ',inp.dimension()
```

The output I get is:

```
/Users/toni_brain/Projects/git/PyEMMA/pyemma/coordinates/util/stat.py:31: DeprecationWarning: Call to deprecated function hist. Called from pyemma.coordinates.util.stat line 31. Please use pyemma.coordinates.histogram()
  def hist(transform, dimensions, nbins):
2.0.1-18-gf8aa02c
  File "minimal_example.py", line 15, in <module>
    inp = coor.source(traj_list, feat)
  File "/Users/toni_brain/Projects/git/PyEMMA/pyemma/coordinates/api.py", line 347, in source
    reader = _create_file_reader(inp, top, features, chunk_size=chunk_size if chunk_size else 100)
  File "/Users/toni_brain/Projects/git/PyEMMA/pyemma/coordinates/data/util/reader_utils.py", line 92, in create_file_reader
    chunksize=chunk_size)
  File "/Users/toni_brain/Projects/git/PyEMMA/pyemma/coordinates/data/feature_reader.py", line 107, in __init__
    self.__set_dimensions_and_lengths()
  File "/Users/toni_brain/Projects/git/PyEMMA/pyemma/coordinates/data/feature_reader.py", line 116, in __set_dimensions_and_lengths
    self._lengths.append(TrajectoryInfoCache[traj])
  File "/Users/toni_brain/Projects/git/PyEMMA/pyemma/coordinates/data/traj_info_cache.py", line 70, in __getitem__
    result = self.__determine_len(filename)
  File "/Users/toni_brain/Projects/git/PyEMMA/pyemma/coordinates/data/traj_info_cache.py", line 76, in __determine_len
    return len(fh)
  File "/Users/toni_brain/miniconda/lib/python2.7/site-packages/mdtraj/formats/xyzfile.py", line 386, in __len__
    raise NotImplementedError()
NotImplementedError
```

A workaround that works is, to load the .xyz file into mdtraj and then convert it to .xtc, save that file and read this into pyemma instead:

```
#Now instead we could load this into mdtraj and then convert to xtc.
#If this xtc is then read into pyemma everything works fine.
import mdtraj as md
traj = md.load('chain.xyz', top='adp.pdb')
traj.save_xtc('chain_mod.xtc')
```

I have attached the .pdb and .xyz files. (renamed with extension .txt for both cases)
[adp.txt](https://github.com/markovmodel/PyEMMA/files/30111/adp.txt)
[chain.txt](https://github.com/markovmodel/PyEMMA/files/30110/chain.txt)

marscher:
Fixes #621

Shows a EfficiencyWarning because it reads the whole file (once by default, because the result will be cached).

### DIFF
--- a/pyemma/coordinates/data/feature_reader.py
+++ b/pyemma/coordinates/data/feature_reader.py
@@ -29,7 +29,6 @@ from pyemma import config
 __author__ = 'noe, marscher'
 __all__ = ['FeatureReader']
 
-
 class FeatureReader(ReaderInterface):
 
     """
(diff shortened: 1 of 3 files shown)


franknoe:
OK, that is a functional workaround, but I still feel that it's weird to do this in PyEMMA. We could suggest this to the mdtraj guys and see if they wanna include it.

marscher:
We can cache the result between runs - mdtraj can not. A better solution would be to just count the frames in the reader (not parsing anything) in mdtraj and cache that result.

franknoe:
OK, looks good. BTW, can we fix the coveralls?

PR #626, opened 2015-11-10
[coor/data/numpy_filereader] new argument dims=[-1]

gph82:
This allows for creating npy_readears for which columns can be chosen at the moment of "chunk-extraction". See the lines

   https://github.com/gph82/PyEMMA/commit/3d13dcea273c4f851138f1a0c9280635bdb1ba00/#L186
   https://github.com/gph82/PyEMMA/commit/3d13dcea273c4f851138f1a0c9280635bdb1ba00/#L192
   https://github.com/gph82/PyEMMA/commit/3d13dcea273c4f851138f1a0c9280635bdb1ba00/#L228
   https://github.com/gph82/PyEMMA/commit/3d13dcea273c4f851138f1a0c9280635bdb1ba00/#L230

@marscher , I took the most convoluted test and replicated it no problem. Do you think that I need to replicate all tests with the new parameter?

### DIFF
--- a/pyemma/coordinates/data/numpy_filereader.py
+++ b/pyemma/coordinates/data/numpy_filereader.py
@@ -67,9 +67,10 @@ class NumPyFileReader(ReaderInterface, ProgressReporter):
         self._array = None
 
         # bookkeeping for dimensions
+        self._ndim_full = None
         if not _is_iterable_of_int(usecols) and usecols is not None:
                 raise ValueError("usecols has to be an iterable of integers! "
-                                 "Instead, dims = %s" % usecols)
+                                 "Instead, usecols = %s" % usecols)
         self._usecols = usecols
 
         self.__set_dimensions_and_lenghts()
@@ -142,17 +143,39 @@ class NumPyFileReader(ReaderInterface, ProgressReporter):
             raise ValueError("input data has different dimensions!"
                              "Dimensions are = %s" % ndims)
 
-        self._ndim = ndims[0]
-        if self._usecols is None:
-            self._usecols = np.arange(self._ndim)
-        else:
-            if np.max(self._usecols) > self._ndim:
-                raise ValueError("Cannot ask for dimension %u, "
-                                 "only %u are available"%(np.max(self._usecols), self._ndim))
-            else:
-                self._ndim=len(self._usecols)
+        #store the original number of dimensions, to  use with the usecol.setter
+        self._ndim_full = ndims[0]
 
+        if self._usecols is None:
+            self._usecols = np.arange(self._ndim_full)
+        else:
+            if np.max(self._usecols) > self._ndim_full:
+                raise ValueError("Cannot ask for dimension %u, "
+                                 "only %u are available"%(np.max(self._usecols), self._ndim_full))
+
+        self._ndim=len(self._usecols)
         self._ntraj = len(self._filenames)
+
+    @property
+    def usecols(self):
+        return self._usecols
+
+    @usecols.setter
+    def usecols(self, newcols):
+        if not _is_iterable_of_int(newcols) and newcols is not None:
+                raise ValueError("usecols has to be an iterable of integers! "
+                                 "Instead, usecols = %s" % newcols)
+        if np.max(newcols) > self._ndim_full:
+                raise ValueError("Cannot ask for dimension %u, "
+                                 "only %u are available"%(np.max(newcols), self._ndim_full))
+
+        # Make sure the setter can restore the full reader
+        if newcols is None:
+            newcols = np.arange(self._ndim_full)
+
+        # If the above sanity checks passed, we're good to go
+        self._usecols = newcols
+        self._ndim = len(newcols)
 
     def _next_chunk(self, context=None):
 
--- a/pyemma/coordinates/data/numpy_filereader.py
+++ b/pyemma/coordinates/data/numpy_filereader.py
@@ -28,7 +28,7 @@ import functools
 
 from pyemma._base.progress import ProgressReporter
 from pyemma.coordinates.data.interface import ReaderInterface
-
+from pyemma.util.types import is_iterable_of_int as _is_iterable_of_int
 
 class NumPyFileReader(ReaderInterface, ProgressReporter):
 
@@ -41,11 +41,15 @@ class NumPyFileReader(ReaderInterface, ProgressReporter):
     chunksize : int
         how many rows are read at once
 
+    usecols : iterable of integers, default = None
+        specifies the columns to be extracted from npy files. If left to None
+        all columns will be extracted
+
     mmap_mode : str (optional), default='r'
         binary NumPy arrays are being memory mapped using this flag.
     """
 
-    def __init__(self, filenames, chunksize=1000, mmap_mode='r'):
+    def __init__(self, filenames, chunksize=1000, mmap_mode='r', usecols=None):
         super(NumPyFileReader, self).__init__(chunksize=chunksize)
 
         if not isinstance(filenames, (list, tuple)):
@@ -61,6 +65,12 @@ class NumPyFileReader(ReaderInterface, ProgressReporter):
 
         # currently opened array
         self._array = None
+
+        # bookkeeping for dimensions
+        if not _is_iterable_of_int(usecols) and usecols is not None:
+                raise ValueError("usecols has to be an iterable of integers! "
+                                 "Instead, dims = %s" % usecols)
+        self._usecols = usecols
 
         self.__set_dimensions_and_lenghts()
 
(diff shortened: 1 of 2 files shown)


gph82:
Here is an example (and a comparison with the other reader):

snap7

marscher:
[-1] is confusing in my opinion, default should be None, which is also a valid slice argument for all elements.

marscher:
I think the test case is ok. Negative column indices are allowed too, right?

gph82:
Yes! 

I'll update to "None" later. Otherwise OK? This will make my life much easier! 

---

## Sent from my phone. Apologies for typos or brevity.

marscher:
Yes, it is a nice feature! Thank you!

franknoe:
Is the numpy reader the only reader where selecting columns makes sense? If not, can we make this as general as possible?

I would suggest to replace "dims" by "columns".

marscher:
It makes sense for all readers besides the feature reader, maybe even there. So this can go to the reader interface

gph82:
Hi @marscher , you find "None" less confusing for "all" columns? I was following the tica's signature:

``` python
tica(data=None, lag=10, dim=-1, var_cutoff=0.95, kinetic_map=True, stride=1, force_eigenvalues_le_one=False, mean=None)
```

where dim=-1 means "I'll pick the dimensions for you". Another option would be dims='all' (or columns='all'), but then we have to make more case differentiation (see #624). What do you guys think?

franknoe:
I would follow this signature:

http://docs.scipy.org/doc/numpy/reference/generated/numpy.loadtxt.html

Am 11/11/15 um 10:44 schrieb Guillermo Pérez-Hernández:

> Hi @marscher https://github.com/marscher , you find "None" less 
> confusing for "all" columns? I was following the tica's signature:
> 
> tica(data=None,lag=10,dim=-1,var_cutoff=0.95,kinetic_map=True,stride=1,force_eigenvalues_le_one=False,mean=None)
> 
> where dim=-1 means "I'll pick the dimensions for you". Another option 
> would be dims='all' (or columns='all'), but then we have to make more 
> case differentiation (see #624 
> https://github.com/markovmodel/PyEMMA/pull/624). What do you guys 
> think?
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/626#issuecomment-155717498.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

franknoe:
PS: I guess the -1 idea follows pythons array selection logic where -1 
is the last element. It's potentially more confusing than None, but it's 
not a big deal and we should stick to it for now to keep the API functional.

Am 11/11/15 um 10:44 schrieb Guillermo Pérez-Hernández:

> Hi @marscher https://github.com/marscher , you find "None" less 
> confusing for "all" columns? I was following the tica's signature:
> 
> tica(data=None,lag=10,dim=-1,var_cutoff=0.95,kinetic_map=True,stride=1,force_eigenvalues_le_one=False,mean=None)
> 
> where dim=-1 means "I'll pick the dimensions for you". Another option 
> would be dims='all' (or columns='all'), but then we have to make more 
> case differentiation (see #624 
> https://github.com/markovmodel/PyEMMA/pull/624). What do you guys 
> think?
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/626#issuecomment-155717498.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
oops, just updated the branch to usecols=None

franknoe:
that's good!

Am 11/11/15 um 10:59 schrieb Guillermo Pérez-Hernández:

> oops, just updated the branch to usecols=None
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/626#issuecomment-155720232.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
BTW, I am a bit unsure about the identity-equality comparison: Isn't this what we want:

``` python
print [1,3,5] is None
print np.array([1,3,5]) is None
print (1,3,5) is None
False
False
False
```

franknoe:
BTW, in the docstring it is still "dims". The protected variable _dims should also be renames to _usecols to avoid developer confusion. Finally, we should clarify first if this column selection should be generalized and if so, how before merging this and create possible create a situation that is harder to generalize

gph82:
just committed that change. I'll just squash the commits and leave this here. Don't know about the best way to generalizing to the ReaderInterface, but looking this PR  might help

franknoe:
Thanks. @marscher and @clonker should have a look at this.

Am 11/11/15 um 11:06 schrieb Guillermo Pérez-Hernández:

> just committed that change. I'll just squash the commits and leave 
> this here. Don't know about the best way to generalizing to the 
> ReaderInterface, but looking this PR might help
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/626#issuecomment-155721444.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

fabian-paul:
Here is a bit of a side comment: @clonker and I recently discussed a similar thing: Why not make dims work like stride? That is to add an extra parameter "dims" next to stride which is handeled in a similar way.
For example the stride parameter can be changed from one pass over the data to the next. We could do the same with dims. This would allow to reimplement the sparisfier in a nice way (features that are excluded by the sparsifier are later not even computed by the featurizer.) And it would allow to implement the Nyström method.

gph82:
@fabian-paul great idea! Can't say I was thinking about Nyström, but, in essence, that's my point (dims~stride) This'll allow for much more flexibility

franknoe:
Am 11/11/15 um 11:18 schrieb fabian-paul:

> Here is a bit of a side comment: @clonker https://github.com/clonker 
> and I recently discussed a similar thing: Why not make dims work like 
> stride? That is to add an extra parameter "dims" next to stride which 
> is handeled in a similar way.
> 
> For example the stride parameter can be changed from one pass over the 
> data to the next. We could do the same with dims. This would allow to 
> reimplement the sparisfier in a nice way (features that are not 
> excluded by the sparsifier are later not even computed by the 
> featurizer.) And it would allow to implement the Nyström method.
> 
> So the main difference to this PR is that your usecols ("dims" -> 
> "usecols") can dynamically change, right?

This makes a lot of sense. But for this usage to be efficient, it will 
require that usecols is used deeply by the reader, i.e. only the 
selected columns should be loaded or featurized. The current PR 
implementation, for example, does not do that because it still loads the 
full array and only passes on part of it (I guess np.load doesn't 
support strided loading ... wtf?). If you do this multiple times, you'd 
always pay the full I/O cost. Similar for the MDFeaturizer: For some 
features the main cost is in computing the features rather than in the 
downstream processing algos, i.e. a selection would need to make sure 
that only the requested features are computed in this pass.

So, all my thumbs are up for going in this direction, but it's not a 
minor change and we have to plan it well.

> —
> https://github.com/markovmodel/PyEMMA/pull/626#issuecomment-155725475.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
@fabian-paul has a good idea for a setter to make it even more dynamical (at the same I/O cost, tough). Indeed this needs more design-meetings, but we'll keep working on a functioning prototype to show what it is we mean

gph82:
Ok, so here's what I have in mind as to how the reader could/should work

tpt11

franknoe:
BTW, array slicing costs something. In your implementation, the default 
has become from simply passing the full array to passing an array that 
has all columns selected. That is significantly slower - which probably 
won't show in this case because I/O is yet slower - but creates an 
unnecessary copy of the array. So for the default selection "None", 
slicing should be completely avoided. You should do something like this 
instead:

``` python
if cols is None:
     return X
else:
     return X[:, cols]
```

Am 11/11/15 um 13:35 schrieb Guillermo Pérez-Hernández:

> Ok, so here's what I have in mind as to how the reader could/should work
> 
> tpt11 
> https://cloud.githubusercontent.com/assets/7518004/11091256/fc5dbcce-8878-11e5-9131-93598403ff93.png
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/626#issuecomment-155765507.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
"BTW, array slicing costs something" Did not know this, thanks.

However, the idea is that usecols get's as deep as possible in the self._next_chunk structure anyway.
I myself am not even sure that the slicing should be done at the return or elsehere, but that's another question.

franknoe:
OK, so can we close this PR and replace it by an issue?

Am 11/11/15 um 19:47 schrieb Guillermo Pérez-Hernández:

> "BTW, array slicing costs something" Did not know this, thanks.
> 
> However, the idea is that usecols get's as deep as possible in the 
> self._next_chunk structure anyway.
> I myself am not even sure that the slicing should be done at the 
> return or elsehere, but that's another question.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/626#issuecomment-155874417.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

gph82:
sure, do you what you feel it's best

franknoe:
OK. Can you create an issue with your suggestion how the reader should work?

gph82:
yupp, no problem

PR #632, opened 2015-11-19
Fix inspect getargspec python35 deprecation

marscher:
fixes 
`DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead`

### DIFF
--- a/pyemma/_base/model.py
+++ b/pyemma/_base/model.py
@@ -18,11 +18,11 @@
 
 from __future__ import absolute_import
 import numpy as _np
-import inspect
+#import inspect
 import warnings
 
 from pyemma.util.statistics import confidence_interval
-from pyemma.util.reflection import call_member
+from pyemma.util.reflection import call_member, getargspec_no_self
 
 __author__ = 'noe'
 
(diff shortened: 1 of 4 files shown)


franknoe:
Hm, is this change fully backwards compatible?
Since the sklearn base class is changed - do we still retain full sklearn compatibility with it?

marscher:
The backported code was written within the Django project to address this problem and has been adopted by SciPy, so yes I guess clever people have written and reviewed it. However the Scikit project offered a much more complicated (and therefore more harder to maintain) solution for the same issue. 
So I've chosen the simpler one... 

https://github.com/scipy/scipy/pull/5248/files 
vs 
https://github.com/scikit-learn/scikit-learn/issues/5281 2 PR, 1 introducing the change and already 1 follow up

franknoe:
Thanks. Since the behavior of the BaseEstimator is unchanged, we should 
still be compatible with sklearn.

Am 20/11/15 um 01:55 schrieb Martin K. Scherer:

> The backported code was written within the Django project to address 
> this problem and has been adopted by SciPy, so yes I guess clever 
> people has written and reviewed it. However the Scikit project offered 
> a much more complicated (and therefore more harder to maintain) 
> solution for the same issue.
> So I've chosen the simpler one...
> 
> https://github.com/scipy/scipy/pull/5248/files
> vs
> scikit-learn/scikit-learn#5281 
> https://github.com/scikit-learn/scikit-learn/issues/5281 2 PR, 1 
> introducing the change and already 1 follow up
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/632#issuecomment-158246497.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

PR #633, opened 2015-11-20
[coor/pca] remove memory optimization (temporaries for sum, dot)

### LINKED ISSUE #631: [coor/pca] when using NumpyFileReader
Hi, I am observing this behavior...seems like Xm and self._dot_prod_tmp have different dimensions?
snap10

marscher:
Since it fails if chunks are smaller than expected.Fixes #631

### DIFF
--- a/pyemma/coordinates/transform/pca.py
+++ b/pyemma/coordinates/transform/pca.py
@@ -76,7 +76,6 @@ class PCA(Transformer):
         default_var_cutoff = get_default_args(self.__init__)['var_cutoff']
         if dim != -1 and var_cutoff != default_var_cutoff:
             raise ValueError('Trying to set both the number of dimension and the subspace variance. Use either or.')
-        self._dot_prod_tmp = None
         self.Y = None
         self._N_mean = 0
         self._N_cov = 0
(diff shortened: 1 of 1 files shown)


franknoe:
@fabian-paul please review.

fabian-paul:
Ok, for some reason setting the out parameter in numpy.dot is not compatible with memaps. Looks like a numpy bug.
The code (that was reverted to the unoptimized version) is ok.

gph82:
Thanks for quick revision! If I understand correctly, the memmap option is intact, it was the optimization that was not compatible with it, right?

I kinda need this bugfix for the weekend, so, if we can merge this soon...

fabian-paul:
Yes, it seems to be a numpy problem. Who knows where else this can pop up. 
But since we get no other error messages, I think our code is fine now.

This is a pure numpy example with the same problem:

``` Python
import numpy as np
np.save('file.npy',np.random.rand(10,10))
f = np.load('file.npy',mmap_mode='r')
a = np.empty(shape=f.shape) 
f.dot(f,out=a)
```

franknoe:
ok, we can merge as soon as tests have passed. But who knows how long AppVeyor will need

marscher:
Appveyor should have finished 10 hours ago...

marscher:
Thanks for the precision fix @fabian-paul. How did you commit to the PR? Didn't know it is possible to do so.

fabian-paul:
Well I just commited to the branch. Because it is in the Markovmodel repo, I have access to it.

marscher:
Ah never mind. I thought this lied on my fork.

fabian-paul:
BTW, if I replace np.empty(shape=f.shape) in the example above by np.empty_like(f), it works. I think that's because the resulting array is then a memmap as well. This is kind of strange, it's a memmap not linked to a file...

marscher:
Maybe it creates a temporary file for that.

fabian-paul:
Or it's an anonymous memmap?
https://en.wikipedia.org/wiki/Mmap#File-backed_and_anonymous
In computer science, everything one can imagine exists.

PR #635, opened 2015-11-25
Fix mlmsm revpi

trendelkampschroer:
This pull-request fixes an inconsistency and a bug. Exposes sampling with fixed stationary vector functionality in the API and exposes an additional keyword argument in two API-functions.

### DIFF
--- a/pyemma/msm/api.py
+++ b/pyemma/msm/api.py
@@ -265,6 +265,7 @@ def markov_model(P, dt_model='1 step'):
 
 
 def estimate_markov_model(dtrajs, lag, reversible=True, statdist=None,
+                          count_mode='sliding',
                           sparse=False, connectivity='largest',
                           dt_traj='1 step', maxiter=1000000, maxerr=1e-8):
     r""" Estimates a Markov model from discrete trajectories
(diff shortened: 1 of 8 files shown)


franknoe:
Thank you! Looks mostly good, I just have two comments:
- Why do you need to expose the counting mode in estimate_markov_model? Is there any reason not to use sliding here?
- Can the tests for BMSM with fixed pi be integrated with the existing BMSM test code? In the ML-MSM test code we treat various cases with almost the same code, which is more powerful and extensible and writing separate test classes for different cases - I imagine it could be the same for BMSMs.

franknoe:
PS: I did on purpose not mirror all parameters of the Estimator class in the api function. The api function should be mainly convenient and easy to understand - it's a design decision what is offered there. If some untypical parameter combinations, they could be accessible using the Object approach

```
mlmsm = MaximumLikelihoodMSM(special_parameter_friend=-327)
mlmsm.estimate(dtrajs)
```

trendelkampschroer:
i) I exposed counting mode in estimate_markov_model so that I can avoid direct construction of the objects in the two notebooks and use the same counting mode for MLE and Bayesian. I also think that it doesn't harm to expose this optional keyword in the API. If I don't need it than I don't have to worry about it, because there is a sensible default, but if I want to change the default I can do it on a single line. I think it does not complicate the API-function  to expose it, but if you disagree I can remove it and go the create object + call estimate by hand route.

ii) Integrating the BMSM-test with fixed pi and the standard BMSM code into a single class would require to completely rewrite the test. Copy and paste and changing some lines in the constructor was much more efficient than rewriting a whole class.

franknoe:
i) OK
ii) Please check if it's possible to combine it. In the MLMSM-Test we treat four cases (reversible/nonreversible, with pi/without pi) with the same code. Of course it's simpler for now to just copy and modify the code, but this is not a sustainable solution. At some point someone else would have to clean that up.

trendelkampschroer:
ii) The test cases are combined now. 

There is a bug in util/statistics, https://github.com/markovmodel/msmtools/issues/66,
that needs to be fixed before the tests can pass.

franknoe:
thanks

Am 26/11/15 um 17:59 schrieb Benjamin Trendelkamp-Schroer:

> ii) The test cases are combined now.
> 
> There is a bug in util/statistics, markovmodel/msmtools#66 
> https://github.com/markovmodel/msmtools/issues/66,
> that needs to be fixed before the tests can pass.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/635#issuecomment-159960168.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

franknoe:
@trendelkampschroer , if my msmtools fix is fine for you, you can merge this. I am traveling tomorrow (friday) an will be unavailable the whole day.

PR #644, opened 2015-12-07
[DTRAM] adding initial files

marscher:


franknoe:
Please don't merge this into devel yet. I haven't had a chance to review the design and naming at all, and as Chris and I discussed we should do that first before leaving the thermo branch.

There will be definitely some naming changes.

franknoe:
You could push this to a new branch that I can edit, but on the other hand refactorings might affect not only the WHAM+TRAM estimators, so I would rather do this in thermo.

PR #651, opened 2015-12-11
[coor refactoring] Fix regspace clustering

marscher:


marscher:
note that this is a work in progress PR for a refactoring branch. The tests for regspace clustering are now all passing.

For backward compat I've added a deprecated "parametrize" method to Transformer, which just calls estimate.

marscher:
@clonker for Kmeans we'll need to know when we're on the last_chunk during initializing the centers. Do you think it could be feasible to add another flag to iterables, so they return a dict of meta information (itraj, timestep, last_chunk), which can the be used in estimators?

clonker:
I think we can provide these just as (maybe in a meta map or configuration object like the iterator context?) members in the iterators themselves.. so something along the lines

```
it = X.iterator()
for chunk in it:
    t = it.timestep
    last_chunk = it.last_chunk
```

marscher:
This looks practical (members).

PR #673, opened 2016-01-14
Split transformer and added notification framework

marscher:
- Transformer is now a slim interface. The pipeline transformer is called StreamingTransformer.
- The notification framework comes with a easy to use decorator ("inform_children_upon_change"), which should be placed on attributes relevant for children of the current class.

We can discuss, if this new interface needs to be improved further.

### DIFF
--- a/pyemma/coordinates/clustering/interface.py
+++ b/pyemma/coordinates/clustering/interface.py
@@ -29,7 +29,7 @@ import os
 from pyemma._base.model import Model
 from pyemma._ext.sklearn.base import ClusterMixin
 from pyemma.coordinates.clustering import regspatial
-from pyemma.coordinates.transform.transformer import Transformer
+from pyemma.coordinates.transform.transformer import StreamingTransformer
 from pyemma.util.discrete_trajectories import index_states, sample_indexes_by_state
 from pyemma.util.files import mkdir_p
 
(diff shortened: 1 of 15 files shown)


franknoe:
Thanks!

I think the Estimator/Transformer relation is still not really clean, but this is as clean as sklearn wants it to be (In sklearn language almost everything is an Estimator, and a Transformer is a specialized estimator that assumes that there is a fit - in reality a Transformer doesn't need to be estimated). So this is something which is fine for now. We can break it up later if needed.

I am puzzled by the fact that the order of inheritance can break functionality in the StreamingTransformer. 
It seems to indicate that we have two different methods with equal names in the tree, and it one of the versions work for us, that probably means we just don't notice what is broken in this version (because we just test for one version of the janus method)

What exactly is broken there?

marscher:
I've removed the DataStreamingElement again to solve the janus problem. Now the interface looks a bit cleaner.

franknoe:
What was the problem?

Am 15/01/16 um 17:02 schrieb Martin K. Scherer:

> I've removed the DataStreamingElement again to solve the janus 
> problem. Now the interface looks a bit cleaner.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/673#issuecomment-172000759.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
Am 15.01.2016 um 17:03 schrieb Frank Noe:

> What was the problem?
> Like you said, a property has been defined two times in the hierarchy 
> (in DataStreamElement and DataSource, which both used to be interfaces 
> of StreamingTransformer).

PR #678, opened 2016-01-21
refactor random access

clonker:
Some performance optimizations and removal of random access for not supported file formats / readers.


marscher:
Thank you. What is the state this PR is based on?

clonker:
The current state (this morning) of the devel_refactor_transformer branch.

marscher:
so why would this merge include the whole prior history of devel_refactor_transformer?

clonker:
You are right, I somehow confused something but now it should be the correct state.

marscher:
Looks good

PR #685, opened 2016-02-12
CSV writer removed to method in DataSource

marscher:


pyemma/coordinates/data/_base/iterable.py
```diff
@@ -73,7 +73,7 @@ def in_memory(self, op_in_mem):
             self._in_memory = op_in_mem
 
     def _clear_in_memory(self):
-        if __debug__:
+        if self._logger_is_active(10):
```
marscher, line 76: replace with logging.DEBUG

pyemma/coordinates/data/_base/iterable.py
```diff
@@ -156,6 +155,54 @@ def get_output(self, dimensions=slice(0, None), stride=1, skip=0, chunk=0):
 
         return trajs
 
+    def write_to_csv(self, filename, stride=1, chunksize=100, **kw):
```
marscher, line 158: make filenames optional (obtain from datasource once #687 is merged)

Issue #695, opened 2016-02-19
progress/bar/gui.py: dummy' object has no attribute 'description'


gph82:
So I get the following error message in one of our tutorial notebooks. Can post the output of conda list package-versions if needed 

``` python
pyemma.__version__
2.0.4-251-gace8547-dirty
...
Y = tica_obj.get_output()[0]
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-12-0442833a7ed8> in <module>()
      3 # here we get the data that has been projected onto the first 2 IC's. It's a list, because we could generally
      4 # have a list of trajectories, so we just get the first element.
----> 5 Y = tica_obj.get_output()[0]
      6 print('Projected data shape = ',Y.shape)

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/transform/transformer.py in get_output(self, dimensions, stride, skip, chunk)
    205             self.estimate(self.data_producer, stride=stride)
    206 
--> 207         return super(StreamingTransformer, self).get_output(dimensions, stride, skip, chunk)
    208 
    209     #@deprecated("Please use estimate")

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/_base/iterable.py in get_output(self, dimensions, stride, skip, chunk)
    152 
    153                 # update progress
--> 154                 self._progress_update(1, stage=1)
    155 
    156         return trajs

/home/mi/gph82/programs/PyEmma/pyemma/_base/progress/reporter.py in _progress_update(self, numerator_increment, stage)
    141             raise Exception("This should not happen")
    142 
--> 143         _show_progressbar(pg)
    144         if hasattr(self, '_prog_rep_callbacks'):
    145             for callback in self._prog_rep_callbacks[stage]:

/home/mi/gph82/programs/PyEmma/pyemma/_base/progress/bar/gui.py in show_progressbar(bar, show_eta)
    157 
    158             # update widgets slider value and description text
--> 159             desc = bar.description
    160             if show_eta:
    161                 desc += ':\tETA:' + bar._generate_eta(bar._eta.eta_seconds)

AttributeError: 'dummy' object has no attribute 'description'
```

gph82:
confirmed that this solved the issue

PR #696, opened 2016-02-19
Fix use of dummy progressbar

jhprinz:
#695 was due to using a dummy progress bar. I added a simple instance check, that the ipython widget is only run if the progress bar is realy a ProgressBar instance. That should fix this bug.


jhprinz:
adding the description to the dummy is (I think) not necessary, but still...

marscher:
Thank you!

PR #722, opened 2016-03-07
[coor.PCA|TICA] implemented partial_fit. Fixes #682

### LINKED ISSUE #682: TICA partial fit
Hello, I was wondering if you could make a [`partial_fit`](http://scikit-learn.org/stable/modules/scaling_strategies.html#incremental-learning) method for the `TICA` class (or if it exists already). So that I can feed it trajectory projections in a for loop and fit it incrementally, without having to give the whole dataset at once.

marscher:
There is one thing I'm currently not sure about. For both PCA and TICA there is a running_covar instance created with n_save estimated from the first data set/array being added. If later on a much larger array is added, we could in principle "need" or want more storages being used, right? 

Should we add a setter for the size of underlying storages?

### DIFF
--- a/pyemma/_base/logging.py
+++ b/pyemma/_base/logging.py
@@ -22,8 +22,7 @@ Created on 30.08.2015
 @author: marscher
 '''
 from __future__ import absolute_import
-
-from logging import getLogger
+import logging
 import weakref
 from itertools import count
 
@@ -35,6 +34,12 @@ class Loggable(object):
     __ids = count(0)
     # holds weak references to instances of this, to clean up logger instances.
     __refs = {}
+
+    _loglevel_DEBUG = logging.DEBUG
+    _loglevel_INFO = logging.INFO
+    _loglevel_WARN = logging.WARN
+    _loglevel_ERROR = logging.ERROR
+    _loglevel_CRITICAL = logging.CRITICAL
 
     @property
     def name(self):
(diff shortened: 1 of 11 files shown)


marscher:
It is not obvious to me, why this one test (which do not use partial_fit) failed...

franknoe:
- That test failure probably means that the "normal" TICA behavior has been changed, which must definitely be avoided.
- In l. 360 of the TICA test, I don't think that
  `np.testing.assert_almost_equal(tica_obj.mean, (means[0] + means[1]) / 2, decimal=2)` is correct. TICA doesn't compute the mean this way, but rather uses 0.5*(X[0:T-tau] + X[tau:T]) for a given X trajectory. In general I would recommend avoiding a "guess" of the right answer at more than one place in the code - TICA estimation is getting more and more sophisticated and the way how the answer is computed depends on several parameters. I suggest to instead do a self consistency check. If you use some n nontrivial random trajectories (not np.random.randn, because these are uncorrelated data, but something which has slow processes), you should get identical results when you run tica on the entire data, or by doing n subsequent partial fits.

franknoe:
BTW, it's worthwhile to check covartools directly, because I don't believe I tested this case. Two scenarios after adding the covariance computer:
1. Add n trajectories, after that compute covar_XX and covar_XY
2. n times, add a trajectory and compute covar_XX and covar_XY
and compare the results

marscher:
Thank you for your remarks. The code now reflects your requirements. I've added some systematic tests and result comparison with a TICA estimate on the whole data set.

Points 1. and 2. are now covered by the TICA tests.

franknoe:
What was the problem leading to the test failure previously?

Am 08/03/16 um 18:01 schrieb Martin K. Scherer:

> Thank you for your remarks. The code now reflects your requirements. 
> I've added some systematic tests and result comparison with a TICA 
> estimate on the whole data set.
> 
> Points 1. and 2. are now covered by the TICA tests.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/722#issuecomment-193868525.

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
Am 08.03.2016 um 18:12 schrieb Frank Noe:

> What was the problem leading to the test failure previously?
> The problem was that the estimation has been performed twice. One time 
> via the API call given the data and another time by calling 
> parametrize(). This led to a different result then, because the 
> RunningMoments instance has been fed twice.

Issue #725, opened 2016-03-09
Handle situations in which logfile can't be written


franknoe:
Importing PyEMMA (at least for the first time, maybe later too...) within a directory that is not writable by the user leads to the following error. This need to be avoided.

```
Traceback (most recent call last):
   File "<stdin>", line 1, in <module>
   File
"/home/mi/noe/miniconda2/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/__init__.py",
line 27, in <module>
     from . import coordinates
   File
"/home/mi/noe/miniconda2/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/coordinates/__init__.py",
line 107, in <module>
     from .api import *
   File
"/home/mi/noe/miniconda2/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/coordinates/api.py",
line 24, in <module>
     from pyemma.util.log import getLogger as _getLogger
   File
"/home/mi/noe/miniconda2/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/util/log.py",
line 110, in <module>
     setupLogging()
   File
"/home/mi/noe/miniconda2/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/util/log.py",
line 93, in setupLogging
     dictConfig(D)
   File "/home/mi/noe/miniconda2/lib/python2.7/logging/config.py", line
794, in dictConfig
     dictConfigClass(config).configure()
   File "/home/mi/noe/miniconda2/lib/python2.7/logging/config.py", line
576, in configure
     '%r: %s' % (name, e))
ValueError: Unable to configure handler 'rotating_files': [Errno 13]
Permission denied: <DIRNAME>
```

marscher:
I'd not consider this a bug, because the user will end up with a non-functional logging system, if we would just ignore the fact that log files can not be written. Maybe we should change the default location of the logfile from current working directory to ~/.pyemma/pyemma.log?

franknoe:
Why is this not a bug? It crashes!

Am 09/03/16 um 15:20 schrieb Martin K. Scherer:

> I'd not consider this a bug, because the user will end up with a 
> non-functional logging system, if we would just ignore the fact that 
[...]

PR #726, opened 2016-03-09
Nicer config handling and wrapping

marscher:
- introduced proper types for config vars
- progressbar handling during tests does not depend on execution order any more
- logging is initialized in the config wrapper
- Fix issue with non-writeable logfiles (fall-back to $HOME/.pyemma/pyemma.log) #725


franknoe:
Awesome, thanks!

PR #727, opened 2016-03-09
Improve import time

marscher:
- by avoiding import lifts
- doing as much imports lazy (when an API function is called)

I'm not sure if we're going to break peoples code by doing so, but anyway users should call the api functions directly instead of importing implementation details.

### DIFF
--- a/pyemma/coordinates/api.py
+++ b/pyemma/coordinates/api.py
@@ -21,41 +21,17 @@ r"""User-API for the pyemma.coordinates package
 
 .. currentmodule:: pyemma.coordinates.api
 """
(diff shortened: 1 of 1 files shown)


franknoe:
Lazy imports are fine, but I am clearly against avoiding the lifts. They were done deliberately, i.e. having easy access to objects are part of the design. I disagree that everything should be done by api functions, they are just a convenience option very much like in sklearn.

marscher:
As you suggested, I rebased out the commit of import lifting removal.

franknoe:
ok!

marscher:
https://pypi.python.org/pypi/Importing/1.10 this can be interesting for us!

PR #728, opened 2016-03-10
Plot its new optargs nits ans process

gph82:
More flexibility when plotting its objects without having to alter the its object itself: 
- Keyword nits  will stop plotting after the nits-th timescale, and avoid cluttering the lower part graph if one estimated many its that are not necessarily needed in the plot
- Keyword process allows for direct selection of the wanted processes, like "show just its nr. 2"  or nrs. [0, 2, 4]. 
- Both keywords are mutually exclusive and work well the other options like showing the refs, the errorbars, and the dt units.

The name of the keywords are chosen to be the same as in its.get_timescale(process=None) and pyemma.msm.its(dtrajs, lag, nits=5). 

I have also added a minimal test, since the code starts to be a bit more complex now. So far, it only tests that things work together without breaking, but is intended to be more comprehensive with time

What do you guys think?

### DIFF
--- a/pyemma/plots/tests/test_its.py
+++ b/pyemma/plots/tests/test_its.py
@@ -1,0 +1,61 @@
+
+# This file is part of PyEMMA.
+#
+# Copyright (c) 2015, 2014 Computational Molecular Biology Group, Freie Universitaet Berlin (GER)
+#
+# PyEMMA is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


marscher:
Thank you for that. Looks like a nice improvement to me.

gph82:
Tests are failing for py 2.X because of slicing the range iterator...since the number of timescales is usually not that high,  _np.arange(ITS.number_of_timescales)[:nits] instead of

```
its_idx = range(ITS.number_of_timescales)[:nits]
```

shouldn't be too harmful, right?

gph82:
One of the appveyor tests fails with:
FAIL: test_partial_fit (pyemma.coordinates.tests.test_featurereader_and_tica.TestFeatureReaderAndTICA)

marscher:
This is caused by a test I wrote recently, which is yet too sensitive. Thanks

PR #734, opened 2016-03-17
[feat] periodic=True fully documented

gph82:
also minor fix here https://github.com/markovmodel/PyEMMA/compare/devel...gph82:distances_opt_arg_periodic_documented?expand=1#diff-d9dec7ccea2883b857a1a78279944970R327


franknoe:
That's fine as a doc fix. We should still add a sanity check too, because this problem is really mean.

gph82:
actually, now that I think about it, don't merge yet, because some other key info (contacts are computed using PBC by force when possible) is not yet there, please do not merge yet

franknoe:
is this mergeable now?

gph82:
Yes

PR #740, opened 2016-03-22
[plots.plot_free_energy]: Fixed bug with plot extends and added some …

franknoe:
…functionality.


franknoe:
If passing, this can be merged into the 2.1 release version.

PR #742, opened 2016-03-23
[coor/tica] Check for too short trajs in partial fit mode.

marscher:
Also added a new field _used_data for partial fit mode to count the number
of frames used in calculation of cov.
- Added len method for LaggedIter
- Added a debug logging handler to record log messages.


marscher:
Fixes a bug mentioned by @stefdoerr in #682

Issue #752, opened 2016-03-31
TICA Error: No array given


stefdoerr:
I used to be able to pass to `Tica.fit_transform()` a list of `np.ndarrays`. Did something change?

``` python
    178     def estimate(self, X, **kwargs):
    179         # TODO: X is either Iterable of an array
    180         if not isinstance(X, Iterable):
    181             if isinstance(X, np.ndarray):
    182                 X = DataInMemory(X, self.chunksize)
    183                 self.data_producer = X
    184             else:
--> 185                 raise ValueError("no array given")
    186 
    187         model = None
    188         # run estimation
    189         try:
    190             model = super(StreamingTransformer, self).estimate(X, **kwargs)
    191         except NotConvergedWarning as ncw:
    192             self._logger.info(
    193                 "Presumely finished estimation. Message: %s" % ncw)
    194         # memory mode? Then map all results. Avoid recursion here, if parametrization
    195         # is triggered from get_output
    196         if self.in_memory and not self._mapping_to_mem_active:
    197             self._map_to_memory()
    198 
    199         self._estimated = True
    200 
    201         return model
    202 

ipdb> not isinstance(X, Iterable)
True
ipdb> type(X)
<class 'list'>
ipdb> type(X[0])
<class 'numpy.ndarray'>
ipdb> X[0].shape
(1200, 595)
```

franknoe:
That is supposed to work. Is this the devel version?

@clonker/@marscher: please comment. If this is not tested, please add a 
test.

Am 31/03/16 um 15:14 schrieb Stefan:

> I used to be able to pass to |Tica.fit_transform()| a list of 
> |np.ndarrays|. Did something change?
> 
> ```
>  178      def  estimate(self,X,**kwargs):
>  179          # TODO: X is either Iterable of an array
>  180          if  not  isinstance(X, Iterable):
>  181              if  isinstance(X, np.ndarray):
>  182                  X  =  DataInMemory(X,self.chunksize)
>  183                  self.data_producer=  X
>  184              else:
> ```
> 
> -->  185                  raise  ValueError("no array given")
>      186  
>      187          model=  None
>      188          # run estimation
>      189          try:
>      190              model=  super(StreamingTransformer,self).estimate(X,**kwargs)
>      191          except  NotConvergedWarningas  ncw:
>      192              self._logger.info(
>      193                  "Presumely finished estimation. Message: %s"  %  ncw)
[...]

PR #753, opened 2016-03-31
fit transform for list of ndarrays

clonker:
Changed the transformer to allow a list of ndarrays upon estimate call and added a test for that.

Also, changed the sklearn/base fit_transform implementation, such that transform gets called on self, rather on what estimate returns, as the returned value might be a model (that does not have a transform method). Or am I mixing something up here and the model should indeed have a transform method?

This should deal with issue #752.

### DIFF
--- a/pyemma/_base/estimator.py
+++ b/pyemma/_base/estimator.py
@@ -127,9 +127,8 @@ def _estimate_param_scan_worker(estimator, params, X, evaluate, evaluate_args,
 
     """
     # run estimation
-    model = None
     try:  # catch any exception
-        model = estimator.estimate(X, **params)
+        estimator.estimate(X, **params)
     except:
         e = sys.exc_info()[0]
         if failfast:
@@ -142,7 +141,7 @@ def _estimate_param_scan_worker(estimator, params, X, evaluate, evaluate_args,
 
     # deal with result
     if evaluate is None:  # we want full models
-        res.append(model)
+        res.append(estimator.model)
     # we want to evaluate function(s) of the model
     elif _types.is_iterable(evaluate):
         values = []  # the function values the model
@@ -155,7 +154,7 @@ def _estimate_param_scan_worker(estimator, params, X, evaluate, evaluate_args,
             # evaluate
             try:
                 # try calling method/property/attribute
-                value = _call_member(model, name, args=args)
+                value = _call_member(estimator.model, name, args=args)
             # couldn't find method/property/attribute
             except AttributeError as e:
                 if failfast:
@@ -334,8 +333,8 @@ class Estimator(_BaseEstimator, Loggable):
 
         Returns
         -------
-        model : object
-            The estimated model.
+        estimator : object
+            The estimated estimator with the model being available.
 
         """
         # set params
@@ -343,7 +342,7 @@ class Estimator(_BaseEstimator, Loggable):
             self.set_params(**params)
         self._model = self._estimate(X)
         self._estimated = True
-        return self._model
+        return self
 
     def _estimate(self, X):
         raise NotImplementedError(
@@ -359,11 +358,12 @@ class Estimator(_BaseEstimator, Loggable):
 
         Returns
         -------
-        model : object
-            The estimated model.
+        estimator : object
+            The estimator (self) with estimated model.
 
         """
-        return self.estimate(X)
+        self.estimate(X)
+        return self
 
     @property
     def model(self):
--- a/pyemma/_ext/sklearn/base.py
+++ b/pyemma/_ext/sklearn/base.py
@@ -432,10 +432,12 @@ class TransformerMixin(object):
         # method is possible for a given clustering algorithm
         if y is None:
             # fit method of arity 1 (unsupervised transformation)
-            return self.fit(X, **fit_params).transform(X)
+            self.fit(X, **fit_params)
+            return self.transform(X)
         else:
             # fit method of arity 2 (supervised transformation)
-            return self.fit(X, y, **fit_params).transform(X)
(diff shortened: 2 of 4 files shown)


franknoe:
I'll have a look. In principle we shouldn't touch sklearn/base, this is 
just a copy of the sklearn base class to ensure compatibility with 
sklearn without creating a package dependency.

Am 31/03/16 um 17:06 schrieb Moritz Hoffmann:

> Changed the transformer to allow a list of ndarrays upon estimate call 
> and added a test for that.
> 
> Also, changed the sklearn/base fit_transform implementation, such that 
> transform gets called on self, rather on what estimate returns, as the 
> returned value might be a model (that does not have a transform 
> method). Or am I mixing something up here and the model should indeed 
> have a transform method?
> 
> This should deal with issue #752 
> https://github.com/markovmodel/PyEMMA/issues/752.
> 
> ---
> 
> ```
>     You can view, comment on, or merge this pull request online at:
> ```
> 
> https://github.com/markovmodel/PyEMMA/pull/753
> 
> ```
>     Commit Summary
> ```
> - [transformer] allow lists of ndarrays in estimate
>   
>   ```
>   File Changes
>   ```
> - _M_ pyemma/_ext/sklearn/base.py
>   https://github.com/markovmodel/PyEMMA/pull/753/files#diff-0 (6)
> - _M_ pyemma/coordinates/tests/test_tica.py
>   https://github.com/markovmodel/PyEMMA/pull/753/files#diff-1 (9)
> - _M_ pyemma/coordinates/transform/transformer.py
>   https://github.com/markovmodel/PyEMMA/pull/753/files#diff-2 (5)
>   
>   ```
>   Patch Links:
>   ```
> - https://github.com/markovmodel/PyEMMA/pull/753.patch
> - https://github.com/markovmodel/PyEMMA/pull/753.diff
> 
> —
> You are receiving this because you are subscribed to this thread.
> https://github.com/markovmodel/PyEMMA/pull/753

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

clonker:
Alright, thanks!

franknoe:
Since `self.fit(X, **fit_params).transform(X)` is the sklearn behavior, our code should implement that in order to be compatible. In terms of the Model/Estimator/Transformer nomenclature I think that a transformer is a specific model, so an estimator that estimates a transformer, should also return (at least) a transformer, and thus the return value of it's fit function is expect to implement `transform()`.

So far for generics. I haven't gone through the code yet to see if there is a problem with that.

franknoe:
@clonker , any idea why only one test is triggered here?

clonker:
That makes sense, I will revert the changes on sklearn/base. :-)
The only reason I made them in the first place however, is that at least the tica model being returned by estimate() does not provide a transform() method, leading to an exception when calling fit_transform.

As for the tests, im not sure why they don't get triggered.. in travis the settings are correctly set to build pull requests. Perhaps it is a problem with travis itself.

franknoe:
I think there is a problem with the Estimator/Model concept in the current code. The estimate and fit routine of an estimator should return self (i.e. an estimator), with the model being fitted and available. For example, in TICA the properties `cov` and `cov_tau` will be available after the fit. However the following example should work too:

``` python
tica = TICA(lag=10, dim=5).fit(X)
print tica.lag
```

which requires that the estimator is being returned because lag is a property of the estimator. Same with PCA.

Conceptually, Estimators are the base. An estimator fit should always return the estimator. Having Models as separate objects is something optional which may have some advantages, e.g. when sampling models (e.g. BayesianMSM). A Transformer should be a subclass of Model, but not of Estimator (there is no reason why a Transformer needs to be estimated).

Transformers and Models can be Mixins that can be added to an Estimator.

Let me know if these explanations are sufficient to adapt the code. If not, it may be easier to meet in person tomorrow or so.

clonker:
So I think I have changed the relevant parts, but please correct me if I'm wrong.
- The `estimate` method in the msm part of pyemma does return only `self`, whereas the it returned `self.model` in the coordinates package. I changed that to `self`. 
- Concerning the Transformers, I have removed that the coordinates base Transformer extends Estimator, however the StreamingTransformer still does (since it first needs to get estimated and then can transform data).

clonker:
By the way: Travis seems to build ([here](https://travis-ci.org/markovmodel/PyEMMA/builds/120027768)), however it does not pop up as pending check in the PR.. odd..

franknoe:
Am 01/04/16 um 11:58 schrieb Moritz Hoffmann:

> So I think I have changed the relevant parts, but please correct me if 
> I'm wrong.
> - The |estimate| method in the msm part of pyemma does return only
>   |self|, whereas the it returned |self.model| in the coordinates
>   package. I changed that to |self|.
> 
> correct.
> - Concerning the Transformers, I have removed that the coordinates
>   base Transformer extends Estimator, however the
>   StreamingTransformer still does (since it first needs to get
>   estimated and then can transform data).
> 
> That's right. The StreamingTransformer is an Estimator, the 
> "Streaming..." part of the name indicates that (stream from data). But 
> the name is not very self-explanatory. If anyone has an idea for a 
> self-explanatory name that is not very long, I'd be happy to hear about it.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/753#issuecomment-204335674

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

franknoe:
Thanks! Merging now.

franknoe:
Fixes #752

Issue #759, opened 2016-04-05
Some difficulties with minRMSD clustering: chunk size shape mismatch


jchodera:
I'm not sure if this is a silly mistake on my part or an issue with the chunking, but I've run into an odd error in minRMSD clustering of a large dataset:

```
[chodera@mskcc-ln1 ~]$ cat cluster-CK2.o7065714
Warning: no access to tty (Bad file descriptor).
Thus no job control in this shell.
Mon Apr  4 17:10:36 EDT 2016
loading reference topology...
Initializing featurizer...
There are 988951 frames total in 557 trajectories.
Clustering...
Traceback (most recent call last):
  File "cluster.py", line 51, in <module>
    dtrajs = clustering.dtrajs
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/coordinates/clustering/interface.py", line 73, in dtrajs
    self._dtrajs = self.assign(stride=1)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/coordinates/clustering/interface.py", line 180, in assign
    mapped = self.get_output(stride=stride)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-2.0.4-py2.7-linux-x86_64.egg/pyemma/coordinates/transform/transformer.py", line 784, in get_output
    trajs[itraj][t:t + L, :] = chunk[:, dimensions]
ValueError: could not broadcast input array from shape (100,1) into shape (60,1)
```

Here is the simple clustering script:

``` python
#!/usr/bin/env python

import pyemma
import numpy as np
import mdtraj
import os

# Source directory
source_directory = '/cbio/jclab/projects/fah/fah-data/munged-with-time/no-solvent/11406' # CK2

################################################################################
# Load reference topology
################################################################################

print ('loading reference topology...')
reference_pdb_filename = 'protein.pdb'
reference_trajectory = os.path.join(source_directory, 'run0-clone0.h5')
traj = mdtraj.load(reference_trajectory)
traj[0].save_pdb(reference_pdb_filename)

################################################################################
# Initialize featurizer
################################################################################

print('Initializing featurizer...')
import pyemma.coordinates
featurizer = pyemma.coordinates.featurizer(reference_pdb_filename)
[...]

PR #770, opened 2016-04-11
Parallel cluster assignment

marscher:
ping @jchodera


jchodera:
Woot! Should I test it out? How do I use it?

marscher:
As it is implemented now, there is no user interaction needed. The process will start as many threads as there are CPUs/cores available. One can restrict this number by setting n_jobs in the API call.

jchodera:
Thanks!

Many cluster environments don't permit this behavior: You are only
allocated so many processes, and using more will stall the job or get it
killed. Is there a way to set the default via an environment variable (much
like OpenMM allows the default number of CPU threads to be selected) if no
number is specified via the API?

marscher:
I can add a probe for the environment variable, which will override the setting the of the API variable.

jchodera:
I think the best approach is to have the environment variable override the default, but the API parameter still take precedence.

marscher:
Isn't that a contradiction? If "OMP_NUM_THREADS" is set, it will always win. For which API values do you suggest to override the env var?

marscher:
Now I got it. Only if the default value in the API is changed, it should take precedence. This makes more sense and we gain more flexibility.

jchodera:
Yes!

marscher:
@jchodera It is working now as expected. Do you want to give it a try? I've tested py27 and py35 locally. Currently conda-build is broken for py27 builds (for whatever reason it does not build anything).

jchodera:
> Currently conda-build is broken for py27 builds (for whatever reason it does not build anything).

Huh.  Anything I can help fix?

jchodera:
> @jchodera It is working now as expected. Do you want to give it a try? 

I'll try installing directly and testing!  However, I think I might still run into #759 on my dataset.

marscher:
too bad. I'll add a solution to this branch then!

marscher:
done. Happy testing!

jchodera:
Woot!  Testing now, thanks!

jchodera:
Testing with 1, 2, 4, and 8 threads now.

jchodera:
Everything died with 

```
Initializing featurizer...
Obtaining file info: 100% (558/558) [##############################] eta 00:00 -There are 1006751 frames total in 558 trajectories.
Clustering...
12-04-16 15:00:52 pyemma.coordinates.clustering.uniform_time.UniformTimeClustering[0] INFO     number of threads obtained from env variable 'OMP_NUM_THREADS'=1
  File "cluster.py", line 53, in <module>
    clustering = pyemma.coordinates.cluster_uniform_time(data=coordinates_source, k=nstates, stride=stride, metric=metric)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-1.2.2+1016.g6891c2c-py2.7-linux-x86_64.egg/pyemma/coordinates/api.py", line 1362, in cluster_uniform_time
    return _param_stage(data, res, stride=stride, chunk_size=chunk_size)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-1.2.2+1016.g6891c2c-py2.7-linux-x86_64.egg/pyemma/coordinates/api.py", line 818, in _param_stage
    this_stage.estimate(X=input_stage, stride=stride)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-1.2.2+1016.g6891c2c-py2.7-linux-x86_64.egg/pyemma/coordinates/transform/transformer.py", line 188, in estimate
    super(StreamingTransformer, self).estimate(X, **kwargs)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-1.2.2+1016.g6891c2c-py2.7-linux-x86_64.egg/pyemma/_base/estimator.py", line 343, in estimate
    self._model = self._estimate(X)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-1.2.2+1016.g6891c2c-py2.7-linux-x86_64.egg/pyemma/coordinates/clustering/uniform_time.py", line 88, in _estimate
    self.clustercenters = np.concatenate([X for X in it])
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-1.2.2+1016.g6891c2c-py2.7-linux-x86_64.egg/pyemma/coordinates/data/_base/datasource.py", line 637, in __exit__
    self.close()
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/pyEMMA-1.2.2+1016.g6891c2c-py2.7-linux-x86_64.egg/pyemma/coordinates/data/feature_reader.py", line 334, in close
    self._logger.debug('closing current trajectory "%s"' % self._data_source.filenames[self.current_trajindex])
IndexError: list index out of range
```

marscher:
Am 12.04.2016 um 23:15 schrieb John Chodera:

> Everything died with
> 
> % self._data_source.filenames[self.current_trajindex]) IndexError: list
> index out of range |

What a shame... and I guess you did not even observe a speedup since you 
set 'OMP_NUM_THREADS'=1. We should really fix this trajindex issue first...

jchodera:
It sadly died before any of the tests got to the timing report for
clustering.

marscher:
@clonker even when we use self.current_trajindex in FeatureReader.close to index the filename list, we get an index error. Can you please have a look?

clonker:
Well i think it is just a missing

``` python
if self._itraj < len(self._data_source.filenames) - 1
```

before the log statement and/or a

``` python
if not self._mditer._closed
```

before the actual closing, as the self.close() that makes trouble gets called by the context `__exit__` in the data source (where it might happen that `_itraj == ntraj`).

franknoe:
Superficial comment: since such failed file open/close attempts were 
leading to multiple issues, could we do things more robustly (i.e. 
without raising exceptions) by doing such file operations always through 
generic helper functions such as tryopen / tryclose which attempt these 
operations carefully and provoke the developer to deal with a failure in 
a non-fatal way?

In general, exceptions are not a big deal when using PyEMMA 
interactively and if the user can avoid them. But for automatization and 
processing of large datasets it's really important that the program is 
robust and fault-tolerant, and doesn't just stop running if there is a 
minor hickup.

Am 13/04/16 um 16:11 schrieb Moritz Hoffmann:

> Well i think it is just a missing
> 
> if  self._itraj<  len(self._data_source.filenames)-  1
> 
> before the log statement and/or a
> 
> if  not  self._mditer._closed
> 
> before the actual closing, as the self.close() that makes trouble gets 
> called by the context |**exit**| in the data source (where it might 
> happen that |_itraj == ntraj|).
> 
> —
> You are receiving this because you are subscribed to this thread.
> https://github.com/markovmodel/PyEMMA/pull/770#issuecomment-209464339

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

jchodera:
+1 for fault tolerance!

Perhaps there could be a mode where issues are logged instead of causing
the fatal exceptions?

I realize this may actually be nontrivial because things like the total
number of snapshots in the dataset may have been determined in one pass or
stage but a file that can't be opened later means fewer snapshots actually
get clustered.

marscher:
An unhandled exception is always fatal and should be fatal since it indicates a serious misbehaviour of the program. This is why the clear answer to this has to be: no this is impossible. Just image writing general purpose exception handlers to catch everything, but how would you handle all possible exceptions in a sane way?

This is why we test our software to avoid as much errors as possible and handle them during runtime (which unfortunately has be be done manually).

marscher:
Of course we could handle broken files non-fatal (which probably will cause silent failures at another spot). It is just a matter of definition how we want to treat these problems (which of course have to be identified before they actually occur).

jchodera:
Let me know when I should give this a try again!

marscher:
I'm still searching for the problem you observe during clustering. The working workaround ha is now on this branch. So you could test again without a crash.

jchodera:
Thanks!  Testing now.

jchodera:
Timing results!

```
chodera@mskcc-ln1 ~]$ cat cluster-CK2-1.o7074258
Warning: no access to tty (Bad file descriptor).
Thus no job control in this shell.
Wed Apr 13 14:37:24 EDT 2016
1
Wed Apr 13 14:37:24 EDT 2016
loading reference topology...
Initializing featurizer...
Obtaining file info: 100% (558/558) [##############################] eta 00:00 -There are 1006751 frames total in 558 trajectories.
Clustering...
13-04-16 14:38:07 pyemma.coordinates.clustering.uniform_time.UniformTimeClustering[0] INFO     number of threads obtained from env variable 'OMP_NUM_THREADS'=1
Elapsed time 1417.023 s

real    25m16.701s
user    17m28.015s
sys 0m26.181s
Wed Apr 13 15:02:41 EDT 2016
[chodera@mskcc-ln1 ~]$ cat cluster-CK2-2.o7074259
Warning: no access to tty (Bad file descriptor).
Thus no job control in this shell.
Wed Apr 13 14:37:24 EDT 2016
2
Wed Apr 13 14:37:24 EDT 2016
loading reference topology...
Initializing featurizer...
Obtaining file info: 100% (558/558) [##############################] eta 00:00 -There are 1006751 frames total in 558 trajectories.
Clustering...
13-04-16 14:38:07 pyemma.coordinates.clustering.uniform_time.UniformTimeClustering[0] INFO     number of threads obtained from env variable 'OMP_NUM_THREADS'=2
Elapsed time 1417.010 s

real    25m15.262s
user    17m48.083s
sys 0m25.882s
Wed Apr 13 15:02:39 EDT 2016
[chodera@mskcc-ln1 ~]$ cat cluster-CK2-4.o7074260
Warning: no access to tty (Bad file descriptor).
Thus no job control in this shell.
Wed Apr 13 14:37:55 EDT 2016
4
Wed Apr 13 14:37:55 EDT 2016
loading reference topology...
Initializing featurizer...
Obtaining file info: 100% (558/558) [##############################] eta 00:01 -There are 1006751 frames total in 558 trajectories.
Clustering...
13-04-16 14:38:35 pyemma.coordinates.clustering.uniform_time.UniformTimeClustering[0] INFO     number of threads obtained from env variable 'OMP_NUM_THREADS'=4
Elapsed time 1767.320 s

real    30m59.181s
user    18m44.929s
sys 0m42.223s
Wed Apr 13 15:08:54 EDT 2016
[chodera@mskcc-ln1 ~]$ cat cluster-CK2-8.o7074261
Warning: no access to tty (Bad file descriptor).
Thus no job control in this shell.
Wed Apr 13 14:37:55 EDT 2016
8
Wed Apr 13 14:37:55 EDT 2016
loading reference topology...
Initializing featurizer...
Obtaining file info: 100% (558/558) [##############################] eta 00:01 -There are 1006751 frames total in 558 trajectories.
Clustering...
13-04-16 14:38:35 pyemma.coordinates.clustering.uniform_time.UniformTimeClustering[0] INFO     number of threads obtained from env variable 'OMP_NUM_THREADS'=8
Elapsed time 1388.790 s

real    24m40.480s
user    15m54.936s
sys 0m27.821s
Wed Apr 13 15:02:36 EDT 2016
```

That's very odd.  8 is a bit faster, but there wasn't much effect from the other numbers of threads.

marscher:
I think the minRMSD metric suffers a lot more from cache misses. The speedups for the euclidean metric are sub-linear. Let me think about this.

franknoe:
Maybe we should think about parallelizing the calculation of distance 
matrices between two sets of structures I and J (e.g. I = the current 
chunk from a trajectory read, J = the current set of clusters centers). 
If caching plays a role here (i.e. if there is a chance to fit multiple 
structures in a cache, which I'm not sure about), we could parallelize 
the calculation of such a distance matrix by blocks.

Anyway it would be good to have input from the mdtraj folks what might 
be going on here. I think the mdtraj rmsd code is highly optimized, and 
maybe we are loosing something here by the way parallel rmsd's are called.

@marscher : how does the parallel RMSD calculation behave on your machine?

Am 14/04/16 um 15:48 schrieb Martin K. Scherer:

> I think the minRMSD metric suffers a lot more from cache misses. The 
> speedups for the euclidean metric are sub-linear. Let me think about this.
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/770#issuecomment-209951526

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

marscher:
Am 14.04.2016 um 17:10 schrieb Frank Noe:

> Maybe we should think about parallelizing the calculation of distance
> matrices between two sets of structures I and J (e.g. I = the current
> chunk from a trajectory read, J = the current set of clusters centers).
> If caching plays a role here (i.e. if there is a chance to fit multiple
> structures in a cache, which I'm not sure about), we could parallelize
> the calculation of such a distance matrix by blocks.
> This seems to be a good solution. I guess we have to test this for a 
> common use case.
> 
> Anyway it would be good to have input from the mdtraj folks what might
> be going on here. I think the mdtraj rmsd code is highly optimized, and
> maybe we are loosing something here by the way parallel rmsd's are called.
> 
> @marscher : how does the parallel RMSD calculation behave on your machine?
> It behaves like what John observed. There is no speedup, it could 
> eventually get even worse.

The pressure on the cache is very high for minRMSD, since we are copying 
both structures to do the in place centering prior distance calculation. 
We could for instance pre-center all cluster centers to avoid this 
re-calculation for all frames.

marscher:
With the latest changes, the minRMSD metric also scales almost linearly. The problem was that the centers were pre-centered within every thread, now it is done for each chunk (which should not dominate for large chunksizes).

We should think about moving this step outside of the assignment step completely and pass pre-centered centers in case of minRMSD metric to the c_extension, of course with a flag set, that there is no need to perform this step).

jchodera:
Woot!  Awesome!  Thanks so much!

Issue #760, opened 2016-04-06
PyEMMA config, logging and trajectory indexing cache


franknoe:
Address issues/policies related to creation of files that log, write config or cache trajectories. Since these issues are closely related, I summarize them here. Feel free to split them into separate issues if more appropriate.
1. **Creation and setting of the PyEMMA configuration file**:
   Optional. By default there is no PyEMMA configuration file and default settings should be used
   (for example, progress bars on if user is in an IPython environment, and otherwise off). Different
   configuration settings will be considered if the user has explicitly created a PyEMMA config file
   under `~/.pyemma/` or has explicitly requested to write that file by something like
   `pyemma.config.save()`. By default, PyEMMA should not try to create a PyEMMA directory or
   write a config file.
   In order to inform the user, the state of the configuration setting should be logged (at a level that is visible by default) when PyEMMA is imported. For example: `PyEMMA v2.1. No configuration file found, defaults are used` or `PyEMMA v2.1. Using configuration found at ~/.pyemma/pyemma.cfg`
2. **Trajectory index cache**:
   By default trajectory caching is done in memory and therefore not permanent / only valid in this interpreter session. If config flag is set _and_ sqllite is available (thread-safeness!), then persistent caching information will be created under `~/.pyemma`. If either of these conditions are not met, or storage creation is unsuccessful, caching information will be be stored persistently during this session. 
   We need a way to clear up the cache. Suggested solution: use a queue for cached entries and maximize the total cache size (e.g. 100 MB).
3. **Logfiles**:
   File Logging is optional and should only be activated by setting a config flag. 

The overall picture is that PyEMMA should only try to create files with useful information if explicitly demanded. Even in this case, creation of files should always be attempted with a fallback option, such that we don't generate crashes when we can't access the disk.
However, in order to make it easy for the average user to change settings, I suggest a global "look and feel" switch such as `pyemma.config.make_my_life_easy(default=False)` (please find better name). If set to false, I suggest reporting that at import time (visibly), such that it's obvious how to activate easy life.

marscher:
A comment to point 2: after reading https://www.sqlite.org/howtocorrupt.html I'm not sure whether it is a good idea to try storing this information on a shared file system.

franknoe:
I'm very open to drop the database / persistent storage for now. If it 
can work robustly, it's nice to have of course. @clonker what do you think?

Am 14/04/16 um 21:23 schrieb Martin K. Scherer:

> A comment to point 2: after reading 
[...]

PR #774, opened 2016-04-15
[config] implemented save/load. Do not try to create a cfg dir by default.

marscher:
Only on invocation of cfg_dir setter or save method files/directories will be
created.

Adresses #760

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -1,5 +1,13 @@
 Changelog
 =========
+2.1.1 (to-be-released)
+----------------------
+
+**Fixes**:
+- In parallel environments (clusters with shared filesystem) there will be no
+  crashes due to the config module, which tried to write files in users home
+  directory. Config files are optional by now.
+
 
 2.1 (29-3-2016)
 ---------------
--- a/pyemma/__init__.py
+++ b/pyemma/__init__.py
@@ -22,6 +22,12 @@ PyEMMA - Emma's Markov Model Algorithms
 =======================================
 """
 from __future__ import absolute_import
+
+# set version from versioneer.
+from ._version import get_versions
+__version__ = get_versions()['version']
+del get_versions
+
 from .util import config
 
 from . import coordinates
@@ -29,10 +35,6 @@ from . import msm
 from . import util
 from . import plots
 from . import thermo
-
-from ._version import get_versions
-__version__ = get_versions()['version']
-del get_versions
 
 
 def setup_package():
--- a/pyemma/coordinates/data/util/traj_info_cache.py
+++ b/pyemma/coordinates/data/util/traj_info_cache.py
@@ -157,6 +157,8 @@ class TrajectoryInfoCache(object):
         return TrajectoryInfoCache._instance
 
     def __init__(self, database_filename=None):
+        # for now we disable traj info cache persistence!
+        database_filename = None
         self.database_filename = database_filename
         if database_filename is not None:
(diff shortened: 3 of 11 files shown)


marscher:
TODO: adopt docstrings.

franknoe:
Please explain in the PR description what this PR does and how it changes behavior. It shouldn't be necessary to read the code in order to know what's in the PR, but anyway I don't understand it even based on reading the code.
Also the documentation needs to be in detail in the doc pages, not just docstrings.
Thanks!

marscher:
Sure. It changes the code from always building the config values first on the default configuration file we ship with the package, then it tries to read existing config files from the hidden directory ".pyemma" in the users home directory. If there is a user-config found, these values will precede the existing ones.

If the user has set a environment variable "PYEMMA_CFG_DIR" to some path, this location will be probed for a configuration directory. If it is not there, the defaults will be used again.

New features:
- save: tries to save the current runtime configuration to the specified configuration directory (which defaults to $HOME/.pyemma/, but can be set via config.cfg_dir = path or os.environ['PYEMMA_CFG_DIR'] = path).
- load: read configuration from a given filename (should be useful for batch scripting purposes).

franknoe:
Sorry, I don't understand:
- It changes the code _from_ ... to what?
- these values will precede the existing ones. What does that mean? Do you mean override the existing ones?

marscher:
Before this PR, pyemma's config module tries to copy over the default files shipped with the package to the users home directory. Now this only happens if the user requests this explicitly by invoking the setter of cfg_dir or by calling save.

Your intuition about the second question is correct, it will just override existing values. The priority is given by (highest: User-cfg, lowest:default-cfg).

marscher:
I will also add a notice in the changelog for 2.1.1

franknoe:
OK, this i my understanding of the PR. Please comment if correct:
- Old behavior
  - If `$HOME/.pyemma/` exists, read the configuration from there and use it.
  - Else: try to create `$HOME/.pyemma/` and copy the default configuration there.
- New behavior
  - If a config is found under PYEMMA_CFG_DIR or `$HOME/.pyemma/`, read the configuration from there and use it. 
  - Else: use default configuration 
  - When calling save, write the current config settings to PYEMMA_CFG_DIR or `$HOME/.pyemma/` for future use.

And all of this happens when calling `import pyemma`. Correct?

marscher:
Am 18.04.2016 um 14:43 schrieb Frank Noe:

> OK, this i my understanding of the PR. Please comment if correct:
> - Old behavior
>     o If |$HOME/.pyemma/| exists, read the configuration from there
>       and use it.
>     o Else: try to create |$HOME/.pyemma/| and copy the default
>       configuration there.
>   Correct.
> - New behavior
>     o If a config is found under PYEMMA_CFG_DIR or |$HOME/.pyemma/|,
>       read the configuration from there and use it.
>     o Else: use default configuration
>     o When calling save, write the current config settings to
>       PYEMMA_CFG_DIR or |$HOME/.pyemma/| for future use.
> 
> And all of this happens when calling |import pyemma|. Correct?
> Precisely.

One question to the "make_my_life_easy" option: I guess the name should 
be something like "show_config_notifications" = True, which by default 
prints instructions how to change/save things, once pyemma is imported.

These notifications could then be turned off by changing this setting.

Would this actually make ones life easier, when we force them to run 
some lines of code just to get rid of messages?

gph82:
trying now to see if #775 is indeed gone...

PR #783, opened 2016-04-25
[plots/timescales] fix off-by-one error in nr. of timescales

### LINKED ISSUE #781: Implied timescale plots fails with a single timescale 2.1
`plots.plot_implied_timescales` seems to have a problem in that it omits one timescale. As a consequence, for a 2-state trajectory that only has a single timescale, nothing will be printed. This problem is new - probably since 2.1. 

As an example, this will plot nothing:

``` python
d = [0, 1, 1, 1, 0, 0, 0, 1, 1, 0, 0, 1, 0, 1, 1, 1, 1, 1, 1]
its = msm.timescales_msm(d, lags = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
plots.plot_implied_timescales(its)
```

Whereas this will plot a single timescale:

``` python
d = [0, 1, 2, 1, 0, 0, 0, 1, 2, 0, 0, 1, 0, 1, 2, 1, 2, 2, 1]
its = msm.timescales_msm(d, lags = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10])
plots.plot_implied_timescales(its)
```

gph82:
fixes #781

### DIFF
--- a/pyemma/plots/timescales.py
+++ b/pyemma/plots/timescales.py
@@ -92,13 +92,15 @@ def plot_implied_timescales(ITS, ax=None, outfile=None, show_mle=True, show_mean
             raise TypeError('optional arguments nits and process are mutually exclusive:', nits, process)
         if not _is_iterable_of_int(process):
             raise ValueError('process has to be an iterable of integers')
-        if _np.max(process) > ITS.number_of_timescales:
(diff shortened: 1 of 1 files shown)


franknoe:
thanks.

Issue #790, opened 2016-04-29
pyemma.coordinates.load suddenly broke


rafwiewiora:
Hi guys!

Not sure what's going on quite yet, but I've been seeing this since last night on `pyemma-dev` and I think `pyemma` conda packages (let me just do a complete clean up of the conda environment in a sec and confirm the release version is also doing this). 

The error:

```
In [2]: x = pyemma.coordinates.load('run0-clone0.h5', top='top.pdb')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-6ccc7bc81304> in <module>()
----> 1 x = pyemma.coordinates.load('run0-clone0.h5', top='top.pdb')

/Users/rafalpwiewiora/anaconda/lib/python2.7/site-packages/pyEMMA-2.1.1-py2.7-macosx-10.5-x86_64.egg/pyemma/coordinates/api.pyc in load(trajfiles, features, top, stride, chunk_size, **kw)
    201                  or len(trajfiles) is 0)):
    202         reader = create_file_reader(trajfiles, top, features, chunk_size=chunk_size, **kw)
--> 203         trajs = reader.get_output(stride=stride)
    204         if len(trajs) == 1:
    205             return trajs[0]

/Users/rafalpwiewiora/anaconda/lib/python2.7/site-packages/pyEMMA-2.1.1-py2.7-macosx-10.5-x86_64.egg/pyemma/coordinates/data/_base/iterable.pyc in get_output(self, dimensions, stride, skip, chunk)
    149                                     description='getting output of %s' % self.__class__.__name__,
    150                                     stage=1)
--> 151             for itraj, chunk in it:
    152                 L = len(chunk)
    153                 if L > 0:

/Users/rafalpwiewiora/anaconda/lib/python2.7/site-packages/pyEMMA-2.1.1-py2.7-macosx-10.5-x86_64.egg/pyemma/coordinates/data/_base/datasource.pyc in next(self)
    621 
    622     def next(self):
--> 623         X = self._it_next()
    624         while X is not None and (
    625                 (not self.return_traj_index and len(X) == 0) or (self.return_traj_index and len(X[1]) == 0)

/Users/rafalpwiewiora/anaconda/lib/python2.7/site-packages/pyEMMA-2.1.1-py2.7-macosx-10.5-x86_64.egg/pyemma/coordinates/data/_base/datasource.pyc in _it_next(self)
    601         self.state.pos = self.state.pos_adv
[...]

PR #791, opened 2016-04-30
Tests of api.load and source for some formats and chunk sizes

clonker:
- Filled an empty code branch, which caused the patches iterator to return None for chunk size 0 and topology formats.
- Added api.load and api.source tests for extensions
  - .binpos
  - .dcd
  - .h5
  - .lh5
  - .nc
  - .netcdf
  - .trr

This should fix issue #790.


franknoe:
Thanks for the super-fast fix @clonker !! I'll merge as soon as this passes.

Two questions out of interest:
- As for multiple formats, what about just storing a single coordinate set (with e.g. 2 frames) and then just specifying a set of formats we want to test against? We could use mdtraj to write a file and then pyemma to retrieve it inside the test - that would then trivially be extensible to all formats that mdtraj can handle. If mdtraj changes the compatibility list in the future, that could automatically adapt if we query the mdtraj supported formats from their API.
- If I understand your fix correctly, you prevent `next(self._mditer)` in `feature_reader` from returning `None`. Then I'm still puzzled by the lines

``` python
    623         X = self._it_next()
    624         while X is not None ...
```

in `datasource`. If X is `None` wouldn't that have caused the same problem in `_it_next()` already?

clonker:
Sure, no problem!
- Yes, i guess it could be done that we create the trajectory files during the test - perhaps using the convert tool or the FormatRegistry of mdtraj.
- Indeed, there was an empty code branch in patches.py, which could lead to returning None values in `next(self._mditer)`. In the lines you pointed out, empty chunks which are not None are skipped. I'm not sure if X can not be None at this point for some corner cases, as datasource contains the parent class for all Reader types.. nevertheless, the test suite runs without errors if I remove the check.

PR #796, opened 2016-05-07
Fixed OpenMP minRMSD clustering.

jchodera:
This PR fixes the issues with minRMSD OpenMP parallel clustering and eliminates the second codepath so that both parallel and serial forms share a single codepath.

I've also taken the suggestion in the code and switched parallelization to be by generator distance computation to minimize cache misses.

### DIFF
--- a/pyemma/coordinates/clustering/src/clustering.c
+++ b/pyemma/coordinates/clustering/src/clustering.c
@@ -120,7 +120,7 @@ int c_assign(float *chunk, float *centers, npy_int32 *dtraj, char* metric,
         ret = ASSIGN_ERR_INVALID_METRIC;
     }
 
-    #pragma omp parallel shared(ret, dim, N_frames, centers, trace_centers_p, dists) default(private)
+    #pragma omp parallel private(buffer_a, buffer_b, i, j, chunk_p, mindist, argmin)
     {
         /* Allocate thread storage */
         buffer_a = malloc(dim*sizeof(float));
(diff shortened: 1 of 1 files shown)


franknoe:
Thanks for the contribution! I think @marscher should review this. At the moment tests seem to be failing with

``` python
Doctest: pyemma.coordinates.api.assign_to_centers ... *** glibc detected *** /home/travis/miniconda/envs/_test/bin/python: free(): invalid next size (fast): 0x000000000192ee70 ***
```

jchodera:
That's odd---it worked locally, and is happily running on our 10^6 snapshot dataset, cutting the clustering time from 7.5 days to 11 hours...

franknoe:
That sounds great! I have no idea what this test failure could be due 
to, but I'm sure it can be fixed then.

Am 07/05/16 um 15:58 schrieb John Chodera:

> That's odd---it worked locally, and is happily running on our 10^6 
> snapshot dataset, cutting the clustering time from 7.5 days to 11 hours...
> 
> —
> https://github.com/markovmodel/PyEMMA/pull/796#issuecomment-217637187

## 

---

Prof. Dr. Frank Noe
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de

jchodera:
[This thread](http://stackoverflow.com/questions/4729395/error-free-invalid-next-size-fast) has some insight.  I had to change exception handling within the OpenMP `parallel` block because a `goto` inside the block that jumps outside wouldn't work properly.

I wonder if it is running into a memory allocation error and then trying to free a pointer that wasn't successfully allocated.  Fixing the travis error may just require checking if the pointer is valid before calling `free`.

franknoe:
Still the same behavior

marscher:
Thank you very much John. The problem is that the dists pointer is only
used in case of minRMSD metric and you have not set this pointer to 0 at
beginning. So in case of Euclidean metric this pointer is not initialized.

Best,
Martin

jchodera:
Thanks! Will check in a fix.

jchodera:
Looks like that fixed it!

franknoe:
yay. Gonna merge it.

PR #797, opened 2016-05-09
Fragmented trajectory reader random access via stride argument

clonker:
Added an implementation of (sorted) random access via the stride argument for the fragmented trajectory reader.


marscher:
Thank you very much!

PR #802, opened 2016-05-13
Refactoring thermo API

cwehmeyer:


franknoe:
Thank you! Minor things:
- Please cite this one for WHAM (you can cite Kumar et al in addition): Ferrenberg AM, Swendsen RH (1989) Optimized Monte Carlo data analysis. Phys. Rev. Lett. 63:1195–1198.
- I am a bit unhappy with plotting functions outside the plots subpackage, because we have always been doubtful about the dependency on matplotlib which as caused problems in the past. Logically pyemma.plots is a higher-level subpackage than the other ones and could in principle be moved out of PyEMMA to a separate package, but then with dependency on PyEMMA. So I would rather create a thermoplots module under pyemma.plots than defining those plots under pyemma.thermo and then importing them into pyemma.plots.

cwehmeyer:
OK, I will take care of this.

franknoe:
OK. Sorry I misunderstood the thermo/plots relation, this is perfectly fine. Ready to merge once the test issues are sorted.

marscher:
We still need to add the notebooks to ipython.rst in doc/source so they will show up on the website.

ping @cwehmeyer

PR #816, opened 2016-05-31
Add state labels to plots.plot_network

franknoe:
Just added a missing parameter to `plots.plot_network`, and removed an uneffective one.


franknoe:
One appveyor test failed with server error, but I can't retrigger the build.

franknoe:
All tests pass, except that on one machine they didn't even start. I think this is mergeable, but let's look again into why I can't retrigger tests on Appveyor.

marscher:
I've just granted our teams the "user" role on Appveyor. Can you please re-try, if you can run builds now?

PR #820, opened 2016-06-01
[conda-recipe] set version by git env variables exported by conda-build

marscher:
This is needed since conda-build will deprecate the way we set the version currently.


franknoe:
ok

PR #829, opened 2016-06-08
[ext/sklearn] updated parameter_search (fixes a undetected test failure in doctest)

marscher:
this version comes from scikit-learn 0.17.1 and makes the ParameterGrid object indexable, which seems to be necessary for the doctest of Estimator.

`TypeError: 'ParameterGrid' object does not support indexing`

Because we are currently using nose to run all tests (including doctests) and nose silently ignores tests in "protected" modules/packages, this one has not yet been caught.


franknoe:
OK

Issue #838, opened 2016-06-23
segfault in KmeansClustering


jchodera:
Has anyone run into segfaults in `KmeansClustering` before?

```
loading reference topology...
Initializing featurizer...
Obtaining file info: 100% (558/558) [##############################] eta 00:01 -There are 696874 frames total in 558 trajectories.
Clustering...
22-06-16 23:19:41 pyemma.coordinates.clustering.kmeans.KmeansClustering[0] INFO     number of threads obtained from env variable 'OMP_NUM_THREADS'=16
initialize kmeans++ centers:   0% (   1/2787) [                    ] eta --:-- \/var/spool/torque/mom_priv/jobs/7516948.hal-sched1.local.SC: line 42: 48885 Segmentation fault      (core dumped) python cluster.py
```

Here is my script:

``` python
#!/usr/bin/env python

import pyemma
import numpy as np
import mdtraj
import time
import os

# Source directory
source_directory = '/cbio/jclab/projects/fah/fah-data/munged3/no-solvent/11406' # CK2
#source_directory = '11406' # CK2

################################################################################
# Load reference topology
################################################################################

print ('loading reference topology...')
reference_pdb_filename = 'protein.pdb'
reference_trajectory = os.path.join(source_directory, 'run0-clone0.h5')
traj = mdtraj.load(reference_trajectory)
traj[0].save_pdb(reference_pdb_filename)

################################################################################
# Initialize featurizer
################################################################################

print('Initializing featurizer...')
import pyemma.coordinates
featurizer = pyemma.coordinates.featurizer(reference_pdb_filename)
#featurizer.add_all() # all atoms
featurizer.add_selection( featurizer.select_Heavy() ) # add heavy atoms

################################################################################
# Define coordinates source
################################################################################

import pyemma.coordinates
from glob import glob
trajectory_filenames = glob(os.path.join(source_directory, 'run*-clone*.h5'))
coordinates_source = pyemma.coordinates.source(trajectory_filenames, features=featurizer)
print("There are %d frames total in %d trajectories." % (coordinates_source.n_frames_total(), coordinates_source.number_of_trajectories()))

################################################################################
# Cluster
################################################################################

print('Clustering...')
generator_ratio = 250
nframes = coordinates_source.n_frames_total()
nstates = int(nframes / generator_ratio)
stride = 1
metric = 'minRMSD'
initial_time = time.time()
[...]

PR #839, opened 2016-06-23
[clustering] define NULL if not defined in clustering header

marscher:
@jchodera this fixes the issue you just observed in kmeans #838. Can you confirm this?


marscher:
note that a recent update of conda-build seems to have broken the win-64 builds... I've already reported upstream.

PR #841, opened 2016-06-23
[coor/api/save_traj] new optarg image_molecules and minor bugfix

gph82:
This incorporates the mdtraj's new feature of making molecules whole (instead of broken across PBCs) to save_traj, given that save_traj's main purpose is visualization

### DIFF
--- a/pyemma/coordinates/api.py
+++ b/pyemma/coordinates/api.py
@@ -545,7 +545,7 @@ def discretizer(reader,
     return disc
 
 
-def save_traj(traj_inp, indexes, outfile, top=None, stride = 1, chunksize=1000, verbose=False):
+def save_traj(traj_inp, indexes, outfile, top=None, stride = 1, chunksize=1000, image_molecules=False, verbose=False):
     r""" Saves a sequence of frames as a single trajectory.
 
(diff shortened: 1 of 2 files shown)


marscher:
Thank you for this improvement! We then also need to require at least 1.7.2 in setup.py and meta.yaml (devtools/conda-recipe) to ensure nothing will break.

marscher:
or probe the Trajectory object if it has the image attribute and trigger a warning instead?

gph82:
I thought we we're doing that already. This can wait, until then (there's ways around it).

marscher:
I think we are enforcing version >=1.7.0, so if this feature has been introduced in 1.7.2, we need to require that too.

gph82:
@marscher , any clever way of re-triggering the appveyor build? Tests are failing because of  "Back-end server is at capacity for url: https://conda.anaconda.org/omnia/noarch/repodata.json.bz2: https://conda.anaconda.org/omnia/noarch/"...so....the intertubes are clogged :)

marscher:
you can either add a (dummy) commit or I can re-trigger the build.

gph82:
I re-triggered the build in travis, but my appveyor account does not allow me to do the same

PR #844, opened 2016-06-27
[coordinates/featurization] fix order of backbone torsions description list

marscher:
In case of backbone torsions we had different orderings for the computed angles and the description list:
- old description: (phi_1, ... phi_n, psi_1, ..., psi_n)
- new: (phi_1, psi_1, ..., phi_n, psi_n)


franknoe:
thanks!

PR #851, opened 2016-07-01
[thermo] fix docstrings of estimate

marscher:


marscher:
just a design question: why did you encapsulate the XX_trajs in a tuple? Isn't it much clearer to have them separately?
@fabian-paul, @cwehmeyer

fabian-paul:
The answer is that estimate() can have only one parameter for the data. The **kwargs is reserved for setting properties.

PR #857, opened 2016-07-05
[coor/feat/angles] cossin=True ignores deg=True

gph82:
Otherwise the user can unknowingly use np.rad2deg on (cos, sin) values , transform to TICA and never notice the error.


marscher:
good catch!

PR #858, opened 2016-07-06
customisable font sizes in networkplots

cwehmeyer:
- the parameter `size` now overwrites the matplotlib `font.size` default (instead of a hard-coded value of 14pt) and sets the label sizes for states and arrows.
- the new parameter `arrow_label_size` is only used  for the arrow labels' font size
- the function `plot_network()` now accepts `**textkwargs` (like `plot_flux()` and `plot_markov_model()`)


franknoe:
thanks

marscher:
Thanks for tidying this up! Appveyor failed for a strange linker error...

PR #890, opened 2016-08-05
Mditer random access uses mdtraj file format read

marscher:
This avoids creating useless Trajectory objects, which eat up a lot of memory. Now the raw data is being collected and a Trajectory object containing the merged data is returned.

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -9,6 +9,8 @@ Changelog
 
 **Fixes**:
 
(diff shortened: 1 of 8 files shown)


clonker:
Your changes already fixed what my changes would have fixed, so they are obsolete.. however I think you need to copy the time attribute as well, right? Hence the test failures...

marscher:
I've discussed the need for the time attribute with Guille when I used 
RA-iterators in frames_from_files and it is pretty useless. So I'll just 
drop the test.

PR #894, opened 2016-08-10
Simulate Trajectories for MSM and HMSM

alexlafleur:
This is our proposal for issue #889.
We implemented a simulate function for both MSM and HMSM to generate trajectories from a given model.

Alexandra and Tara

### DIFF
--- a/pyemma/msm/models/hmsm.py
+++ b/pyemma/msm/models/hmsm.py
@@ -435,3 +435,52 @@ class HMSM(_MSM):
 
         """
         return _np.argmax(self.observation_probabilities, axis=0)
+
+
+    def simulate(self, time_steps, initial_states=None , num_traj=1):
+        """
+        Generates trajectories of the underlying model.
+        There are two possibilities to use this function:
+
+        Either
+            * given parameter num_traj the function returns an amount of (num_traj) trajectories.
+        Or
+            * given parameter initial_states the function return an amount of (initial_states) size trajectory regarding the supplied starting states.
+            The parameter num_traj is ignored.
+        ----------
+        time_steps : int
+           number of time steps to simulate
+        initial_states : int array, optional, default=None
+            set of initial state indices
+        num_traj : int, optional, default = 1
+            number of trajectories
+
+        Returns
+        -------
+        trajectories :
+            An tuple of hidden trajectories and the observables
+        """
+        import scipy.stats
+        if num_traj is not None:
+            M = _MSM(self.P)
+            hidden_trajectories = M.simulate(time_steps=time_steps,num_traj=num_traj)
+
+
+        elif initial_states is not None:
+            M = _MSM(self.P)
+            hidden_trajectories = M.simulate(time_steps=time_steps, initial_states=initial_states)
+
+        else:
+            raise ValueError('Please specify either num_traj or initial_states to simulate a trajectory')
+
+        observables = _np.ndarray(_np.shape(hidden_trajectories), int)
+        kr = _np.shape(observables)[0]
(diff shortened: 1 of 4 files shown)


greenTara:
In the process of writing these extensions, we noted a few opportunities to improve the code, but wanted to discuss this before proceeding.
1. There are differences in naming conventions between the MSM and HMSM class (aliases np and _np for numpy respectively)
2. There is inconsistent documentation in HMSM (attribute P is called Pcoarse, attribute pobs is called Pobs)

franknoe:
I'll take care of reviewing this. However, already a few changes:
- please use identical nomenclature as in msmtools.generation.generate_traj.
- generate one trajectory only with this function, otherwise the return value is unclear.
- never use the rv_discrete function inside a loop, because that's expensive. Pre-generate all needed distributions before the loop, and possible save them to the object, then just sample from them inside the loop.
- There are no tests, both functions need to be tested.

franknoe:
P.S.: hmsm.simulate() is actually not that easy to do, especially if the user wants to specify an observable state rather than a hidden state to start from (that's possible). I have code for this, better leave it to me and just to the msm.simulate binding now.

franknoe:
PPS: by "identical nomenclature as in msmtools.generation.generate_traj" I mean input parameters and names for start, stop, length etc.

franknoe:
I currently don't have enough bandwidth to pull, so I'm just giving an example for a simple HMSM.simulate here. This doesn't have the option to start from observable states. This feature can be added later, but if you wanna give it a shot: If the first observed state is given, then the first hidden state needs to be sampled from the distribution 

```
p(hidden | observed) = p(hidden) p(observed | hidden) // normalized.
```

Here, `p(hidden)` is the stationary distribution of the hidden transition matrix, `p(observed | hidden)` is the output distribution, //normalized indicates normalization to sum 1 by hidden state. That can be written in two lines of matrix operations, but since I don't have the up-to-date code here I won't write it, otherwise I'll just get the variable names wrong.

This code snip is not tested, please adapt if needed and add tests. MSM can be according (that doesn't really need to be more than a line)

``` python
    def simulate(self, N, start=None, stop=None, dt=1):
    """ Generates a realization of the Hidden Markov model.

    Parameters
    ----------
    N : int
        trajectory length in steps of the lag time
    start : int, optional, default = None
        starting hidden state. If not given, will sample from the stationary 
        distribution of the hidden transition matrix.
    stop : int or int-array-like, optional, default = None
        stopping hidden set. If given, the trajectory will be stopped before 
        N steps once a hidden state of the stop set is reached
    dt : int
        trajectory will be saved every dt time steps.
        Internally, the dt'th power of P is taken to ensure a more efficient simulation.

    Returns
    -------
    htraj : (N/dt, ) ndarray
        The hidden state trajectory with length N/dt
    otraj : (N/dt, ) ndarray
        The observable state discrete trajectory with length N/dt

    """
    from scipy import stats
    # generate output distributions
    output_distributions = [stats.rv_discrete(values=(np.arange(pobs.shape[1]), pobs_i)) for pobs_i in pobs]
    # sample hidden trajectory
    htraj = msmtools.generation.generate_traj(self.transition_matrix, N, start=start, stop=stop, dt=dt)
    otraj = np.zeros(htraj.size, dtype=int)
    # for each time step, sample microstate
    for t, h in enumerate(htraj):
        otraj[t] = output_distributions[h].rvs()  # current cluster
    return htraj, otraj
```

alexlafleur:
We assumed that the start parameter would indicate the hidden start state.

franknoe:
Thank you very much.

PR #902, opened 2016-08-18
Enable subset to handle active sets in pyemma.thermo

cwehmeyer:


cwehmeyer:
@marscher do you want to add anything before we merge?

marscher:
well, the new subset is pretty much untested (33% coverage). I thought it should be used by the feature in thermo (which then itself seems untested).

marscher:
The unittest for subset is in _base, which is ignored by nosetest stupidity...

marscher:
nose-devs/nose/issues/963

marscher:
Please do not merge, this is broken...

marscher:
We can merge this now, right? @cwehmeyer

cwehmeyer:
Not yet. We need to address some new issues with this.

PR #904, opened 2016-08-18
[coordinates] add skip parameter to estimators to skip initial n_frames of each file.

marscher:
The interface seems redundant at the first view (eg. for a pipeline one has to give the same skip parameter for each element in the pipeline), ... but it totally fits into the logic of separating the IteratorState from the actual DataSource.

It also makes sense to store skip as estimation parameter, since it behaves like stride: it manipulates how much data we are using.

PS: I hope I have not forgotten to pass any "skip"

### DIFF
--- a/pyemma/coordinates/api.py
+++ b/pyemma/coordinates/api.py
@@ -361,7 +361,7 @@ def source(inp, features=None, top=None, chunk_size=None, skip=0, **kw):
     return reader
 
 
-def pipeline(stages, run=True, stride=1, chunksize=100):
+def pipeline(stages, run=True, stride=1, chunksize=1000):
     r""" Data analysis pipeline.
 
     Constructs a data analysis :class:`Pipeline <pyemma.coordinates.pipelines.Pipeline>` and parametrizes it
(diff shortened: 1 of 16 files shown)


marscher:
@jchodera this is now a much cleaner impl as it does not fiddle around with lengths internally but just use the already present skip parameter. Please note that the in case of a pipeline, you have to use the same skip parameter in every step, otherwise your "faulty" data would be used again!

jchodera:
How does it work now? Previously, we just added `skip` to

``` python
coordinates_source = pyemma.coordinates.source(trajectory_filenames, features=featurizer)
```

to make

``` python
coordinates_source = pyemma.coordinates.source(trajectory_filenames, features=featurizer, skip=nskip)
```

jchodera:
Would I just change

``` python
clustering = pyemma.coordinates.cluster_uniform_time(data=coordinates_source, k=nstates, stride=stride, metric=metric)
```

to

``` python
clustering = pyemma.coordinates.cluster_uniform_time(data=coordinates_source, k=nstates, stride=stride, metric=metric, skip=nskip)
```

?

marscher:
yes, the skip argument has to be passed to the actual step reading the data during estimation.

jchodera:
Thanks!

PR #906, opened 2016-08-22
Fix traj info db if config directory is not set.

marscher:


marscher:
@jchodera I've tested this with the conditions you have on Travis with msm-pipeline.

jchodera:
Thanks!

PR #921, opened 2016-09-06
[estimator] don't use joblib in case of n_jobs=1

marscher:


gph82:
thanks for the speed, @marscher !

marscher:
these random failures are really pissing me off....

PR #929, opened 2016-09-12
fix time units in MFPT in msm.tpt and all related quantities

fabian-paul:
Please review.


franknoe:
That appears to make sense, thanks.

marscher:
Thanks! Could you please add an entry for the changelog (doc/src/CHANGELOG.rst), since this will give users different results for all TPT related quantities?

PR #867, opened 2016-07-12
Models estimators serializable

marscher:
Estimators and Models are now serializable via the derivation of the SerializableMixin class. This class provides a save method and the analogue function to restore a dumped class is pyemma.load(filename).

It is being tried to dump as little information as possible in order to restore the class state. For Estimators the parameters used during estimation will be stored, and also the estimated model, if there is any. A model is also simply serialized by dumping the model parameters, defined by the set_model_params method of the model.
This already handles a lot of use cases, but we still need the ability of storing extra (private) attributes for an estimator to fully obtain the desired runtime behaviour. This is achieved by defining a static field in the class of interest, which just enumerates the extra parameters.

All the impl details are hidden in the SerializableMixin class, so one just needs to define a version number and an optional extra fields list to enable the saving/loading of the class. The version number is used to decide whether we can restore the file trivially (version number unchanged) or need to perform a mapping between the old and new version of the class.
The interpolation mapping handles the following operations: add, remove, move to handle the situation of refactored classes during restoration. The map is capable of dealing with a version history and executes all interpolations in a linear way to mimic the order of changes to the class.
Even the renaming of classes could be handled in a central place (pyemma.load).

Note that this feature is the basis for #322 

Resolves #810

PR #931, opened 2016-09-16
Refactor msm use properties

marscher:
The attributes transition matrix, stationary_distribution used to be getters for P and pi respectively. This has been replaced by single properties, which set derived quantities and the prior long name getters are just aliases for these.

This simplifies the MSM class a bit, which is needed for #867.

It fully preserves the API.

It also adds discards imaginary parts of eigen decompositions, if all imaginary parts are zero, in order to avoid some problems.


franknoe:
looks good for me, thanks.

PR #933, opened 2016-09-19
[coor/save_traj] fix parameter verbose

### LINKED ISSUE #932: pyemma.coordinates.save_coor: obsolete outputs even with verbose=False
Hi,

in the pyemma.coordinates.save_coor method, regardless of whether I set verbose=False or =True, I always get an output statement for each particular coordinate file that I just wrote out, e.g.:

9-09-16 18:26:38 pyemma.coordinates.api INFO     Created file WT.pcca2_100samples.bs_0.dcd
INFO:pyemma.coordinates.api:Created file WT.pcca2_100samples.bs_0.dcd

This can be quite cumbersome for a large number of coordinate files, but the "verbose" parameter should actually control this output behavior, correct?

Thanks,
Sebastian

gph82:
fixes #932

### DIFF
--- a/pyemma/coordinates/api.py
+++ b/pyemma/coordinates/api.py
@@ -549,7 +549,7 @@ def discretizer(reader,
     return disc
 
 
-def save_traj(traj_inp, indexes, outfile, top=None, stride = 1, chunksize=1000, image_molecules=False, verbose=False):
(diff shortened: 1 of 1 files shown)


marscher:
looks good. Thanks!

PR #941, opened 2016-09-22
Ticafixes

stefdoerr:
I cast the numpy array to python int to avoid the warning https://github.com/markovmodel/PyEMMA/issues/939#issuecomment-248850663

I directly cast to the correct dtype in transform to fix `tica.transform()` returning `float64` instead of `float32` like `get_output()`. https://github.com/markovmodel/PyEMMA/issues/939#issuecomment-248847992

That's it! I hope I'm forgiven for my false-issue :smile:

### DIFF
--- a/pyemma/coordinates/data/_base/iterable.py
+++ b/pyemma/coordinates/data/_base/iterable.py
@@ -351,7 +351,7 @@ class LaggedIterator(object):
         return min(n1, n2)
 
     def __len__(self):
-        return int(min(self._it.trajectory_lengths(), self._it_lagged.trajectory_lengths()))
+        return min(self._it.trajectory_lengths().min(), self._it_lagged.trajectory_lengths().min())
 
     def __iter__(self):
         return self
--- a/pyemma/coordinates/data/_base/iterable.py
+++ b/pyemma/coordinates/data/_base/iterable.py
@@ -351,7 +351,7 @@ class LaggedIterator(object):
         return min(n1, n2)
 
     def __len__(self):
-        return min(self._it.trajectory_lengths(), self._it_lagged.trajectory_lengths())
+        return int(min(self._it.trajectory_lengths(), self._it_lagged.trajectory_lengths()))
 
     def __iter__(self):
         return self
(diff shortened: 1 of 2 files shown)


marscher:
Looks good to me.

pyemma/coordinates/data/_base/iterable.py
```diff
@@ -351,7 +351,7 @@ def _n_chunks(self):
         return min(n1, n2)
 
     def __len__(self):
-        return min(self._it.trajectory_lengths(), self._it_lagged.trajectory_lengths())
+        return int(min(self._it.trajectory_lengths(), self._it_lagged.trajectory_lengths()))
```
marscher: actually its better to obtain the minimum of each numpy array by invoking min() on the array and pass this to pythons min

stefdoerr: Yeah I was a bit confused because you seem to do min on two numpy arrays which would not work if they have length more than one.
Changed it now in any case.

marscher: Thank you

stefdoerr: Did this break python 2.7?

stefdoerr: maybe 2.7 requires `np.min(arr)` instead of `arr.min()`? Seemed to pass travis before

stefdoerr: Although the error seems to be in something totally unrelated.

``` pytb
======================================================================

FAIL: test_transition_matrix_samples (pyemma.msm.tests.test_bayesian_msm.TestBMSM)

----------------------------------------------------------------------

  File "/Users/travis/miniconda/envs/_test/lib/python2.7/site-packages/pyEMMA-2.2.5+4.g765cc0d-py2.7-macosx-10.6-x86_64.egg/pyemma/msm/tests/test_bayesian_msm.py", line 83, in test_transition_matrix_samples

    self._transition_matrix_samples(self.bmsm_revpi)

  File "/Users/travis/miniconda/envs/_test/lib/python2.7/site-packages/pyEMMA-2.2.5+4.g765cc0d-py2.7-macosx-10.6-x86_64.egg/pyemma/msm/tests/test_bayesian_msm.py", line 93, in _transition_matrix_samples

    assert msmana.is_reversible(P)

AssertionError
```

marscher: It is unrelated, yes. Stochastic failures

PR #952, opened 2016-10-05
Estimator worker fix evaluate

marscher:
This feature used to be broken. Now the arguments for evaluate are correctly handled. Added some tests also.

six has to be at least 1.10 for "PY34" attribute.


franknoe:
thanks, looks good.

PR #959, opened 2016-10-13
Fix for the fragmented trajectory reader in random access mode

clonker:
Deals with issue #956.


marscher:
Thank you for the quick fix!

PR #963, opened 2016-10-14
[coor/transform] Commute map weighting

maxentile:
Adds an option for tICA to weight eigenvectors following:

[[Noé, Banisch, Clementi, 2016] Commute maps: separating slowly-mixing molecular configurations for kinetic modeling](http://pubs.acs.org/doi/abs/10.1021/acs.jctc.6b00762)


franknoe:
Thanks. kinetic_map and commute_map should be mutually exclusive, i.e. when both set to True it should rather raise, and selection should be done with elif.

franknoe:
Also, I guess it would help to have the option in the coordinates API function too.

franknoe:
Thanks Josh, this looks good!

PR #966, opened 2016-10-17
Remove msmtools api

### LINKED ISSUE #550: remove pyemma.msm.$(msmtools subpackages)
Due to experience in the workshop it is confusing to users, if high-level objects and low-level API mix up in autocomplete (they've chosen to use the low level api, because the naming of desired quantities are more promising).

marscher:
fixes #550


franknoe:
+1

marscher:
Some of the notebooks still used msmtools.io.load_dtrajs, I refactored that by directly importing this function from msmtools in the notebooks.
However I think we should provide a loader at the high-level API

PR #967, opened 2016-10-19
[coordinates] introduced new config var "coordinates_check_output"

marscher:
- The option will trigger the check for inf and NaN in iterator output.
- Can be enabled by config.coordinates_check_input=True. Exception contains position information.

After all the hassle with a faulty mdtraj computation routine (#891) , I'll introduce this for debugging purposes.


marscher:
@cwehmeyer, @fabian-paul If we check for infinity here, could this possibly interfere in the future with streaming of energies in thermo? One could also add individual checks for nan and inf, but at this state I think this solution is sufficient.

PR #969, opened 2016-10-20
Frames from files fixes

### LINKED ISSUE #958: [frames_from_files] ValueError: largest specified index is larger than trajectory length
Getting `ValueError: largest specified index is larger than trajectory length` even with staged trajectories.

@maxentile and I had seen this for non-staged trajectories that we had reason to believe were being continuously updated. This was mentioned previously (https://github.com/markovmodel/PyEMMA/issues/892). More recently this problem has been happening with trajectories that have been copied to a different folder and aren't actively updated.

```
traj = frames_from_files(trajfiles, top, indexes, chunksize, stride, reader=reader)
 File "/cbio/jclab/home/hansons/opt/anaconda/lib/python2.7/site-packages/pyemma/coordinates/data/util/frames_from_file.py", line 111, in frames_from_files
   length))
ValueError: largest specified index (436 * stride=436 * 1=436) is larger than trajectory length '/cbio/jclab/home/hansons/sims/AZ/SYK/11407/run7-clone19.h5' = 240
```

Also, of note, if I try and run the same analysis on just that trajectory, the error does not occur and everything runs fine. There is also so far no evidence that the HDF5 file itself is bad.

Also note pyemma version:

```
conda list| grep pyemma
pyemma                    2.2.6               np110py27_0    omnia
```

### LINKED ISSUE #968: [coor.save_traj] different behaviour dcd and xtc?
An `indexes` vector produced by `cl.sample_indexes_by_cluster([0], 10)` gives the right results when used back on the original, continuous variable, the tic `Y`, meaning that slicing `Y` with `indexes`,  and assigning it to a clustercenter gives...the index of the clustercenter that produced `indexes` in the first place (0 in this case).

However, when using the same `indexes` as input for `coor.save_traj`, the output trajectory objects belong to other clusterscenters, different from 0. 

This behavior:
- occurs with .dcds but NOT with .xtcs (only tested these two so far). 
- goes away if one reduces the number of input features, perhaps hinting at the fact that a poorly defined tica eigenproblem is somewhat behind this

We have issue #771 in connection with .dcds, but i think its unrelated, since the featurization itself yields the same results for dcds and xtcs (see assertion statements in the code). 
However, the tics themselves ARE different (between .dcd and .xtc) 

in the case of poorly defined tica (very few frames, many input features), again hinting at the fact that there is something happening at the transformation level.

Sorry for not posting the code directly, this is just to get the ball rolling...
snap

marscher:
Fixes #968, because of unitcell_lengths scaling was missing.
Fixes #958, because the index checking was faulty.


franknoe:
thanks!

PR #970, opened 2016-10-20
[github] added templates for new issues and pull requests

marscher:


marscher:
When merged, creators of new issues and pull requests will be shown the messages in the .github folder. We're currently lacking the AUTHORS file, I'll add it in a second.

PR #975, opened 2016-10-25
Fix: support for numpy integers in TICA lag

stefdoerr:
Take care when checking against `int`. It doesn't work for numpy ints. There are other solutions than the one with `import numbers`. For example casting to `int` and comparing to original value, but I think this is more intuitive.

``` python
In [18]: isinstance(5, int)
Out[18]: True

In [19]: isinstance(np.int64(5), int)
Out[19]: False

In [20]: isinstance(np.int64(5), numbers.Integral)
Out[20]: True

In [21]: isinstance(5, numbers.Integral)
Out[21]: True
```

### DIFF
--- a/pyemma/coordinates/data/_base/iterable.py
+++ b/pyemma/coordinates/data/_base/iterable.py
@@ -18,6 +18,7 @@
 from abc import ABCMeta, abstractmethod
 import six
 import numpy as np
+import numbers
 
 from pyemma._base.logging import Loggable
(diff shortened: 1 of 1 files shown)


franknoe:
that's correct.

marscher:
actually we should have used the util.types.is_int function, where this case is handled like you just proposed.

stefdoerr:
Ok, I'll change the PR

marscher:
Thank you!

PR #977, opened 2016-10-26
Fix frag ra

### LINKED ISSUE #973: [coor.save_traj] with fragmented trajectory reader
It pains me to open this, but:

``` python
pyemma.__version__
# '2.2.7'
src
# <pyemma.coordinates.data.fragmented_trajectory_reader.FragmentedTrajectoryReader at 0x7f4ebf8dce80> 
isample.shape
# (200, 2)
type(isample)
# numpy.ndarray
pyemma.coordinates.save_traj(src.filenames, isample, None, top=struct, stride=10)
#  <mdtraj.Trajectory with 200 frames, 504 atoms, 28 residues, and unitcells at 0x7f4e8c13dc50>
```

However, if instead of parsing the filenames I parse the `FragmentedTrajectoryReader`:

``` python
pyemma.coordinates.save_traj(src, isample, None, top=struct, stride=10)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-27-09af50b1dc42> in <module>()
----> 1 pyemma.coordinates.save_traj(src, isample, None, top=struct, stride=10)

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/api.py in save_traj(traj_inp, indexes, outfile, top, stride, chunksize, image_molecules, verbose)
    660                          % (len(trajfiles), indexes[:,0].max()))
    661 
--> 662     traj = frames_from_files(trajfiles, top, indexes, chunksize, stride, reader=reader)
    663 
    664     # Avoid broken molecules

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/util/frames_from_file.py in frames_from_files(files, top, frames, chunksize, stride, verbose, copy_not_join, reader)
    132         it = reader.iterator(chunk=chunksize, stride=sorted_inds, return_trajindex=False)
    133         with it:
--> 134             collected_frames = [f for f in it]
    135         dest = _preallocate_empty_trajectory(top, len(frames))
    136         t = 0

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/util/frames_from_file.py in <listcomp>(.0)
    132         it = reader.iterator(chunk=chunksize, stride=sorted_inds, return_trajindex=False)
    133         with it:
--> 134             collected_frames = [f for f in it]
    135         dest = _preallocate_empty_trajectory(top, len(frames))
    136         t = 0

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/_base/datasource.py in __next__(self)
    629 
    630     def __next__(self):
--> 631         return self.next()
    632 
    633     def _use_cols(self, X):

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/_base/datasource.py in next(self)
    667 
    668     def next(self):
--> 669         X = self._it_next()
    670         while X is not None and (
    671                 (not self.return_traj_index and len(X) == 0) or (self.return_traj_index and len(X[1]) == 0)

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/_base/datasource.py in _it_next(self)
    647         self.state.pos = self.state.pos_adv
    648         try:
--> 649             X = self._use_cols(self._next_chunk())
    650         except StopIteration:
    651             self._last_chunk_in_traj = True

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/fragmented_trajectory_reader.py in _next_chunk(self)
    305                 raise StopIteration()
    306 
--> 307         X = next(self._it, None)
    308         if X is None:
    309             raise StopIteration()

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/fragmented_trajectory_reader.py in __next__(self)
    115                         self._reader_it.close()
    116                         if self.ra_indices is not None:
--> 117                             self._reader_it = self._select_next_ra_iterator()
    118                         else:
    119                             self._reader_it = self._readers[self._reader_at].iterator(self._stride, return_trajindex=False)

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/fragmented_trajectory_reader.py in _select_next_ra_iterator(self)
    132         while len(ra_indices) == 0:
    133             self._reader_at += 1
--> 134             ra_indices = self.__get_ifrag_ra_indices(self._fragment_indices, self._reader_at)
    135         return self._readers[self._reader_at].iterator(ra_indices, return_trajindex=False)
    136 

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/fragmented_trajectory_reader.py in __get_ifrag_ra_indices(self, fragment_indices, ifrag)
    212     def __get_ifrag_ra_indices(self, fragment_indices, ifrag):
    213         offset = self._cumulative_lengths[ifrag - 1] if ifrag > 0 else 0
--> 214         ra = self.ra_indices[fragment_indices[ifrag]] - offset
    215         indices = np.zeros((len(ra), 2), dtype=int)
    216         indices[:, 1] = ra.squeeze()

IndexError: list index out of range
```

I can continue using the first form, but it creates a new reader every time. The fragmented trajectory reader is a black-box for me. Any ideas?

marscher:
should fix #973 by catching the condition that reader_at index runs out of bounds.

### DIFF
--- a/pyemma/coordinates/data/_base/datasource.py
+++ b/pyemma/coordinates/data/_base/datasource.py
@@ -24,6 +24,7 @@ import six
 from pyemma.coordinates.data._base.iterable import Iterable
 from pyemma.coordinates.data._base.random_accessible import TrajectoryRandomAccessible
 from pyemma.util import config
+from pyemma.util.annotators import deprecated
 from six import string_types
 import os
 
@@ -385,9 +386,14 @@ class DataSourceIterator(six.with_metaclass(ABCMeta)):
         return self.state.is_stride_sorted()
 
     @property
-    def _n_chunks(self):
+    def n_chunks(self):
         """ rough estimate of how many chunks will be processed """
         return self._data_source.n_chunks(self.chunksize, stride=self.stride, skip=self.skip)
+
+    @property
+    @deprecated("use n_chunks")
+    def _n_chunks(self):
+        return self.n_chunks
 
     def number_of_trajectories(self):
         return self._data_source.number_of_trajectories()
--- a/pyemma/coordinates/data/_base/datasource.py
+++ b/pyemma/coordinates/data/_base/datasource.py
@@ -135,7 +135,7 @@ class DataSource(Iterable, TrajectoryRandomAccessible):
 
                 raise ValueError("Input data has different dimensions ({dims})!"
                                  " Files grouped by dimensions: {groups}".format(dims=res.keys(),
-                                                                                  groups=res))
+                                                                                 groups=res))
 
             self._ndim = ndims[0]
             self._lengths = lengths
@@ -165,14 +165,22 @@ class DataSource(Iterable, TrajectoryRandomAccessible):
         """
         return self
 
-    def number_of_trajectories(self):
+    def number_of_trajectories(self, stride=None):
         r""" Returns the number of trajectories.
+
+        Parameters
+        ----------
+        stride: None (default) or np.ndarray
 
         Returns
         -------
             int : number of trajectories
         """
-        return self._ntraj
+        if not IteratorState.is_uniform_stride(stride):
+            n = len(np.unique(stride[:, 0]))
+        else:
+            n = self._ntraj
+        return n
 
     def trajectory_length(self, itraj, stride=1, skip=None):
         r"""Returns the length of trajectory of the requested index.
@@ -184,6 +192,8 @@ class DataSource(Iterable, TrajectoryRandomAccessible):
         stride : int
             return value is the number of frames in the trajectory when
             running through it with a step size of `stride`.
+        skip: int or None
+            skip n frames.
 
         Returns
         -------
@@ -192,23 +202,27 @@ class DataSource(Iterable, TrajectoryRandomAccessible):
         if itraj >= self._ntraj:
             raise IndexError("given index (%s) exceeds number of data sets (%s)."
                              " Zero based indexing!" % (itraj, self._ntraj))
-        if isinstance(stride, np.ndarray):
+        if not IteratorState.is_uniform_stride(stride):
             selection = stride[stride[:, 0] == itraj][:, 0]
             return 0 if itraj not in selection else len(selection)
         else:
             return (self._lengths[itraj] - (0 if skip is None else skip) - 1) // int(stride) + 1
 
     def n_chunks(self, chunksize, stride=1, skip=0):
-        """ rough estimate of how many chunks will be processed """
+        """ how many chunks an iterator of this sourcde will output, starting (eg. after calling reset())
+
+        Parameters
+        ----------
+        chunksize
+        stride
+        skip
+        """
         if chunksize != 0:
-            if not DataSourceIterator.is_uniform_stride(stride):
-                chunks = ceil(len(stride[:, 0]) / float(chunksize))
-            else:
-                chunks = sum((ceil(l / float(chunksize))
-                              for l in self.trajectory_lengths(stride=stride, skip=skip)))
+            chunks = sum((ceil(l / float(chunksize))
+                          for l in self.trajectory_lengths(stride=stride, skip=skip)))
         else:
-            chunks = self.ntraj
-        return int(chunks)
+            chunks = self.number_of_trajectories(stride)
+        return chunks
 
     def trajectory_lengths(self, stride=1, skip=0):
         r""" Returns the length of each trajectory.
@@ -225,8 +239,9 @@ class DataSource(Iterable, TrajectoryRandomAccessible):
         -------
         array(dtype=int) : containing length of each trajectory
         """
-        n = self.number_of_trajectories()
-        if isinstance(stride, np.ndarray):
+        n = self.ntraj
+
+        if not IteratorState.is_uniform_stride(stride):
             return np.fromiter((self.trajectory_length(itraj, stride)
                                 for itraj in range(n)),
                                dtype=int, count=n)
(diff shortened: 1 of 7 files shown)


gph82:
Great! Thanks for doing this so quickly, I'll get back to you ASAP

clonker:
I believe you are missing out on two frames with this pullrequest, can you (also) try to run it off my branch clonker/fix_frag_ra?

gph82:
Sorry @marscher but I keep getting the same issue as in #973 although i'm using pyemma version
2.2.7+6.g523604f

``` python
amax = indexes[::100][:, 1].argmax()
print(indexes[::100][amax])
print(indexes[::100][amax]*10)
print(src.trajectory_lengths())
#[     0 110350]
#[      0 1103500]
#[1114545  510650]
```

Meaning the frag_rdr should be able to get the largest frame index, 110350 (=1103500 when un-strided back x10 to the original), but it still isn't getting it.

gph82:
Ok, @clonker's version (https://github.com/clonker/PyEMMA/commit/f32a25d8444e6481accd9c0a4a4e5074bb60cdca) passes the test and appears to retrieve the right geometries. 

We'll try to write this in a a working unit test

gph82:
@marscher , while I was discussing with @clonker , we thought about including an (optional?) progress bar for the most time consuming part of `frames_from_file` . However:
- Is it really a good idea?
- Instead of being an option, what about having it appear only in cases where it'll take more than x-minutes?
- What's the performance loss in going from list comprehension to an actual loop here:
  https://github.com/markovmodel/PyEMMA/blob/c7e1f39e2da78d5e161139ee2f8862576fd5a862/pyemma/coordinates/data/util/frames_from_file.py#L134?

My answers are:
- 1&2 I think they're both good ideas. I use heavily save_traj interactively and the times vary
  greatly depending on what exactly lies in `indexes` and what type of reader one has.
- 3) IDK about that: best answer I got so far was this http://stackoverflow.com/a/22108640 saying 
  that it's actually not that bad

I have implemented it the basics here:
https://github.com/marscher/PyEMMA/compare/fix_frag_ra...gph82:martin_fix_frag_ra?expand=1, 
 it seems to work except for the warning:`This should not happen. An caller pretended to have achieved more work than registered`

marscher:
The progress reporter would need some logic to not pop up at all, if the first ETA is below a threshold.
If you get this warning, then you update the PG more often than work steps you've registered before. This indicates that _n_chunks is not correct for the RA iterator.
I don't had a strong opinion about the list comprehension, it is just simpler to read. Of course you can not use it, if a PG is desired.

gph82:
what do you mean "_n_chunks is not correct for the RA iterator", n_chunks is taken from "it"

marscher:
GitHub  Hickup (not forwarding email replies...): The attribute n_chunks of the iterator is smaller than the actual amount of chunks, this why you get this warning. Therefore it is not correct for random access iterators (at least in this case).

gph82:
sorry Martin, I don't understand: what is faulty, the iterator or the progress reporter?

marscher:
On 26.10.2016 18:00, Guillermo Pérez-Hernández wrote:

> sorry Martin, I don't understand: what is faulty, the iterator or the
> progress reporter?

the iterator, because the n_chunks property reports less chunks than the 
actual outcome.

gph82:
So even if the iterator "thinks" he's got more chunks than what he actually has, it is still passing all of our tests and giving the right results? Probably it's not that faulty and it just stops iterating once he's been trough all the strided frames?

marscher:
You still dont understand right? :)

gph82:
Let me tell you what I understand: you have said that the fragmented trajectory reader is faulty. From then on, what's not clear to me is what you think that we should try to do to fix it (given that, as you say, it's faulty)

:)

marscher:
The _n_chunks property of the FragmentedReader needs to be fixed.

marscher:
On 26.10.2016 18:14, Martin Scherer wrote:

> The _n_chunks property of the FragmentedReader needs to be fixed.
> Since this property is derived from DataSource, it is faulty in general 
> for non-uniform strides:

```
             chunks = ceil(len(stride[:, 0]) / float(chunksize))
```

But this looks correct, doesn't it?

clonker:
Odd.. it looks correct to me as well. Anyways - the `chunksize==0` case is not fully correct in the random access case, since trajectories might be omitted. I guess it should be something like

``` python
chunks = len(np.unique(stride[:, 0]))
```

marscher:
You're right, I wasn't aware we allow dupes.

marscher:
added a solution for n_chunks

gph82:
thanks, I can confirm this makes the warning go away

gph82:
Hey can we omit the last commit as not a part of the "fix"?

marscher:
Mhm, I thought you wanted this feature. Of course I can kick it out again, you still have the commit on some other branch, right? PS: since when email replying aint reliable any more, this is really annoying

gph82:
yeah IDK about email, it's weird, I'm sticking to online answers. 

I still have the commit and I want to work it a bit more (time-checking etc) before merging, because in some cases this might produce a lot of unwanted verbosity.

Let's leave it OUT for now. The fra ra is working which was the important thing and I'll see about the progress bar at some other time

marscher:
There should a general solution for the "verbosity problem".
You should adopt https://github.com/marscher/progress_reporter/blob/master/progress_reporter/

I think it would be the best option to check the ETA time in _show_progressbar() and if it is smaller than a sane threshold, never show it.

gph82:
Ok, we'll do that.

marscher:
On 27.10.2016 15:11, Guillermo Pérez-Hernández wrote:

> Hey can we ommit the last commit as not a part of the "fix"?

Mhm, I thought you wanted this feature. Of course I can kick it out 
again, you still have the commit on some other branch, right?

PR #979, opened 2016-10-27
[ci] use nose-randomly plugin to reset random seed for each test

marscher:
This should increase coverage and can also reveal different testing
failures which depend on the execution order.

In general this is a minor change.


marscher:
closing in favour of #980

PR #980, opened 2016-10-31
[CI] switch to pytest to test the package in conda-build

### LINKED ISSUE #743: nose excludes tests in _base package silently
Since there is no way around this, we should transition to pytest soon (as soon as 2.9.4 is released, because of cwd deleted bug).

marscher:
Background is that nosetests is no longer maintained (>4 years).
Pytest is modern and maintained with tons of available plugins.

Enabled support for CircleCI.
Fixed some minor testing issues.

Fixes #743


jchodera:
@marscher: I'm curious what the benefits of CircleCI over travis-ci might be. Can you summarize what your findings are?

marscher:
I'm not intending to replace Travis. One benefit of Circle should be, that it offers parallelism over input, eg. one can split tests across several workers/containers. They offer 4 parallel workers for open source projects, which is the same what Travis provides. For this "benefit" they lack build matrix support, which is essential. But if one want to built a bleeding fast CI for "instant" commit testing, this service seems promising.

jchodera:
Thanks, @marscher!

PR #1020, opened 2017-01-05
[coor/featurizer/angles] fix reshaping

### LINKED ISSUE #1018: issue with add_angle(cossin=True)
I'm trying to add angles with the _cossin=True_ feature for later use with TICA.  With _cossin = False_ I get no problems, with _cossin = True_ I get the following error:

       trajs[itraj][it.pos:it.pos + L, :] = chunk[:, dimensions]
       IndexError: too many indices for array

The problem is that the _FeatureReaderIterator_ object returns a flat (1-D) _chunk_ object when _cossin = True_ rather than the (at least) 2-D object required. 

Attached is a zip file with the conda environment, the PDB topology file, the DCD trajectory file and a script to recreate the issue.  I'm using pyEmma  2.2.7 (via Anaconda) with OS X 10.10.5.  I've tried it in both a python 2.7 and 3.5 environment and I get the same issue. 

Thanks for your time, 

Rob
[angles_bug.zip](https://github.com/markovmodel/PyEMMA/files/685788/angles_bug.zip)

marscher:
Fixes #1018

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -1,6 +1,22 @@
 Changelog
 =========
 
+2.3 (tba)
+---------
+
+**New features**:
+- coordinates:
+  - ...
+  - ...
+
+
+**Fixes**:
+
+- coordinates:
+  - ...
+  - ...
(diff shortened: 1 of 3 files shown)


franknoe:
thx

PR #1023, opened 2017-01-06
Changelog for 2.3

fnueske:
Updated changelog. We will have an arxiv-link for the OOM paper on Monday. We can update the references next week or wait, as you like.


franknoe:
I'm just merging this because it's only docs. We're good go!

PR #1026, opened 2017-01-10
[msm]: Updated timescales_msm() docstring. fixes #1025

### LINKED ISSUE #1025: BayesianMSM / counting modes
In the BayesianMSM estimator, the default counting mode is set to 'effective' which is different to the maximum likelihood estimator default. I understand that counting in sliding window mode underestimates the error, but does it make sense to implicitly use different counting modes? The results can be quite different, so it might be good to have this documented.
Further, is there a reason not to use the effective counting mode in MLE?

thempel:


marscher:
Fixes #1025

marscher:
Congrats to your first PR :tada:

PR #1036, opened 2017-02-07
add Python 3.6 to travis build matrix

mj-harvey:
Adds Python 3.6 to the travis test matrix (3.6 now default in Anaconda release)


marscher:
Our dependencies are not yet on 3.6, so this is can not work at the moment.

PR #1039, opened 2017-02-08
[coor] test partial_fit with different TICA parameters

fabian-paul:


marscher:
The current diff is just equivalent to #1038

PR #1041, opened 2017-02-13
[coor] fix problem with negative regularized implied time scales in commute map

fabian-paul:


franknoe:
thanks

fabian-paul:
Fixes the problem addressed in the lower part of https://github.com/markovmodel/PyEMMA/issues/1037#issuecomment-278212295

PR #1046, opened 2017-02-16
Fix docstring inheritance decorator

marscher:
Some derived attributes were not fixed by the decorator.


marscher:
The failure of the single osx job is unrelated (http issues on anaconda cloud)

PR #1061, opened 2017-03-21
Ducktype weights in Covariance API

marscher:
This change makes it obsolete to inherit a custom weight class from the hidden _Weights class in the koopman implementation. The prior solution is java-style and not user friendly.


marscher:
Addresses #1060

Issue #480, opened 2015-08-03
[source] should be allowed to produce output with very high stride


gph82:
I'm aware that the entire transformation is jut to huge for memory, but with a (ridiculously) high enough stride I should be able to get 4124 frames x 57 dimensions, right?

``` python
print my_SOURCE.dimension()
print my_SOURCE.n_frames_total()
mystride = int(1e3)
print my_SOURCE.n_frames_total()/mystride
#57 ndim
#4124963 n_original frames
#4124 n_frames after stride
```

However:

``` python

D = my_SOURCE.get_output(stride=mystride)

---------------------------------------------------------------------------
MemoryError                               Traceback (most recent call last)
<ipython-input-52-b6755ba4bb48> in <module>()
----> 1 D = my_SOURCE.get_output(stride=mystride)

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/transform/transformer.pyc in get_output(self, dimensions, stride)
    630                                'getting output of ' + self.__class__.__name__)
    631 
--> 632         for itraj, chunk in self.iterator(stride=stride):
    633             if itraj != last_itraj:
    634                 last_itraj = itraj

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/transform/transformer.pyc in next(self)
     60         last_itraj = self._transformer._itraj
     61         if self._lag == 0:
---> 62             X = self._transformer._next_chunk(lag=self._lag, stride=self._stride)
     63             return (last_itraj, X)
     64         else:

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/data/feature_reader.pyc in _next_chunk(self, lag, stride)
    166         :return: a feature mapped vector X, or (X, Y) if lag > 0
    167         """
--> 168         chunk = self._mditer.next()
    169         shape = chunk.xyz.shape
    170 

/home/mi/gph82/programs/PyEmma/pyemma/coordinates/util/patches.pyc in iterload(filename, chunk, **kwargs)
    113             while True:
    114                 if extension not in _TOPOLOGY_EXTS:
--> 115                     traj = f.read_as_traj(topology, n_frames=chunk*stride, stride=stride, atom_indices=atom_indices, **kwargs)
    116                 else:
    117                     traj = f.read_as_traj(n_frames=chunk*stride, stride=stride, atom_indices=atom_indices, **kwargs)

xtc.pyx in xtc.XTCTrajectoryFile.read_as_traj (mdtraj/formats/xtc/xtc.c:3584)()

[...]

PR #1073, opened 2017-04-24
Use random access feature reader high stride

marscher:
see issue #480 for background.

The problem is circumvented by creating a RA stride array to collect the desired frames. If the stride is bigger than a cutoff, this strategy is used and avoids reading a lot of unwanted data. The other active option is that if stride*chunksize*n_atoms exceeds a certain memory limit.

The values are of course debatable, eg. the stride cutoff maybe to large for big systems, since it should be more io efficient to just seek to the frames, if the frame lengths exceeds a certain buffer size used by the operating system (or hard drive).

### DIFF
--- a/pyemma/coordinates/data/util/reader_utils.py
+++ b/pyemma/coordinates/data/util/reader_utils.py
@@ -69,6 +69,7 @@ def create_file_reader(input_files, topology, featurizer, chunk_size=1000, **kw)
                 raise ValueError("The passed list did not exclusively contain strings or was a list of lists "
                                  "(fragmented trajectory).")
 
+        # TODO: this does not handle suffixes like .xyz.gz (rare)
         _, suffix = os.path.splitext(input_list[0])
 
         # check: do all files have the same file type? If not: raise ValueError.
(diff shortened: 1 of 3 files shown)


marscher:
@gph82 It would be nice, if you could run a test with your big data set.

gph82:
Hi @marscher thanks for looking into this. I'll try to test it this week give me some time for updating all other things that were left pending before and during Houston

gph82:
Hi, I recovered original example (this is from the time we started htica with BPTI). I've checked with the current  devel version and this PR....both work! 

The thing is, I had my machine upgraded from 8GB to 16GB RAM in the two years (!) between now and the original issue, so I cannot reproduce the original problem to begin with....

HOWEVER:
2.3.2+17.g24b7a7c.dirty (devel)
timeit -n 5 -r 5  mysrc.get_output(stride=mystride)
5 loops, best of 5: 23.6 s per loop

2.3.2+20.g27ace36.dirty (this PR)
timeit -n 5 -r 5  mysrc.get_output(stride=mystride)
5 loops, best of 5: 3.25 s per loop

which is pretty awesome, if you ask me. 

Also, nothing else seems broken.

marscher:
Thank you for your feedback! I'm going to merge this now.

PR #1080, opened 2017-05-10
Fixed missing import in coordinates/transform/tica

fnueske:
Fixed missing import in coordinates/transform/tica.


marscher:
This was used but not imported? Thank you!

fnueske:
Yeah. I introduced this warning with my last PR, and I must have made some mistake with my git copy. The import was present in my copy, but it was absent in the file that was merged here. I don't really understand it and I hope it is fine now.

PR #1082, opened 2017-05-10
[ML-MSM] towards removing dtrajs attribute (deprecated first)

marscher:
As discussed yesterday, I started to remove the storage of discrete trajectories in the estimator. This makes the class more sklearn-like in term it separates data from the actual estimation.

The properties, which rely on _dtrajs_full now have an equivalent "compute_$attribute". These methods take a dtrajs argument.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,27 +1,15 @@
 import pytest
+import sys
+
+# this is import for measuring coverage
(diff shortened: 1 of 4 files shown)


franknoe:
I'll review it.

PR #1091, opened 2017-05-18
[kmeans] allow to pass the seed into kmeans

marscher:
Now bool and int can be passed to kmeans "fixed_seed" parameter. If fixed_seed is False, then a time based seed is used. If it is True, seed is 42 (like the previous behaviour). In seed is an integer, just ensure it fits into unsigned int (C) and pass it to the extension.

@franknoe This will circumvent your problems with "fast" kmeans

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -26,7 +26,7 @@ Changelog
 
 - coordinates:
    - kmeans: allow the random seed used for initializing the centers to be passed. The prior behaviour
-     was to init the generator by time, if fixed_seed=False. Now bool and int can be passed. #
+     was to init the generator by time, if fixed_seed=False. Now bool and int can be passed. #1091
 
 **Fixes**:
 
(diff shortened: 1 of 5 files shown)


franknoe:
Thanks. Two questions:
- What will happen if you pass 0 or 1? Are these interpreted as bool or as integer?
- Is there time-based seed method that uses milliseconds instead of seconds? That would also solve my problem without having to generate seeds externally. The problem was just the 1 second is too long.

marscher:
Sorry I replied yesterday by email, but thunderbird decided not to use the reply-to field, so that mail got lost.

0 or 1 is interpreted as integer, because the check for bool is the first.

If one includes milliseconds along with the seconds since epoch, eg. like this:
int(time.time()*1000)
one can not pass it into C's srand(unsigned int) because this data type only accepts a maximum  of 2**32-1.

I've changed the seeding to draw a positve integer from [0, 2**32-1], this should add enough randomness for the seed.

franknoe:
ok, I think we can merge and try this.

PR #1097, opened 2017-05-19
New data

cwehmeyer:
Added a generator for multi-ensemble data.

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -29,7 +29,7 @@ Changelog
      was to init the generator by time, if fixed_seed=False. Now bool and int can be passed. #1091
 
 - datasets:
-   - added a multi-ensemble data generator for the 1D asymmetric double well.
+   - added a multi-ensemble data generator for the 1D asymmetric double well. #1097
 
 **Fixes**:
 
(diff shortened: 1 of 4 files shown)


doc/source/CHANGELOG.rst
```diff
@@ -28,6 +28,9 @@ Changelog
    - kmeans: allow the random seed used for initializing the centers to be passed. The prior behaviour
      was to init the generator by time, if fixed_seed=False. Now bool and int can be passed. #1091
 
+- datasets:
+   - added a multi-ensemble data generator for the 1D asymmetric double well.
```
marscher: please add the number of this PR here.

cwehmeyer: done

marscher (changes requested):
Could you please add a simple invocation of this, to see it does not fail? Actually I trust you that you've tested this well, but if somebody is going to change (break) it in the future, we have no chance to catch this early.

Put it in datasets.tests.test_datasets.

marscher (approved):
perfect!

marscher:
I just cancelled the second appveyor job (there is no need to test both the branch and the PR merge to devel in this case).

PR #1100, opened 2017-05-23
[thermo] remove progress bar for 'finding connected set'

### LINKED ISSUE #1098: 'finding connected set' progress bar not properly removed
The progress bar shown for finding the connected set in a TRAM estimation is not properly removed after the estimation terminates.

fabian-paul:
fixes #1098


fabian-paul:
Actually this should be done with some context manager, because we want the progress bars to be removed if there is an exception (or even have have them turn red if there is an exception...)
Maybe like this:
```Python
with stage(self, 'compute_the_answer_to_the_ultimate_question') as s:
    for ...
        # very long calculations
        s.progress_update(1)
# hooray!

marscher:
a context manager is nothing more than syntactic sugar for a try catch finally block. But I'll add this as a feature request to progress reporter ;)

PR #1104, opened 2017-05-24
Fix state labels plot networks

marscher:
plot_flux: If sets A, B contained more than one state, automatic state
labelling did not respect that (only first and last state were called 'A' and 'B' respectively).

NetworkPlot:
- Also the documentation states, that if ax is None, a new figure would
be created, but plt.gcf() was used. Now a proper new figure is created.
- Changed plain asserts to proper exception handling.
- added simple logging (debug).
- added some tests. use only pyemma functionality (removed msmtools refs).

### DIFF
--- a/pyemma/plots/networks.py
+++ b/pyemma/plots/networks.py
@@ -23,8 +23,13 @@ import warnings
 from pyemma.util import types as _types
 from six.moves import range
 from math import sqrt as _sqrt
+import logging
+
+logger = logging.getLogger(__name__)
+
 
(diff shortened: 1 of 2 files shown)


marscher:
@cwehmeyer It would be nice, if you could have a look at my changes, since I'm not sure if the figure creation has implications on the current assumptions.

cwehmeyer (approved):

marscher:
Thank you Chris!

PR #1111, opened 2017-05-30
Amm

psolsson:
Adds AMM functionality. Unit-tests passing.

### DIFF
--- a/pyemma/msm/api.py
+++ b/pyemma/msm/api.py
@@ -27,6 +27,7 @@ from .estimators import BayesianHMSM as _Bayes_HMSM
 from .estimators import MaximumLikelihoodMSM as _ML_MSM
 from .estimators import OOMReweightedMSM as _OOM_MSM
 from .estimators import ImpliedTimescales as _ImpliedTimescales
+from .estimators import AugmentedMarkovModel as _AMM
 
 from .models import MSM
 from pyemma.util.annotators import shortcut
--- a/pyemma/msm/api.py
+++ b/pyemma/msm/api.py
@@ -38,7 +38,7 @@ import numpy as _np
 __docformat__ = "restructuredtext en"
 __author__ = "Benjamin Trendelkamp-Schroer, Martin Scherer, Frank Noe"
 __copyright__ = "Copyright 2014, Computational Molecular Biology Group, FU-Berlin"
-__credits__ = ["Benjamin Trendelkamp-Schroer", "Martin Scherer", "Frank Noe", "Simon Olsson"]
+__credits__ = ["Benjamin Trendelkamp-Schroer", "Martin Scherer", "Frank Noe"]
 __maintainer__ = "Martin Scherer"
 __email__ = "m.scherer AT fu-berlin DOT de"
 
@@ -1307,17 +1307,11 @@ def estimate_augmented_markov_model(dtrajs, ftrajs, m, w, lag,
     dtrajs : list containing ndarrays(dtype=int) or ndarray(n, dtype=int)
         discrete trajectories, stored as integer ndarrays (arbitrary size)
         or a single ndarray for only one trajectory.
-    ftrajs : 
-       
-        
-    m   : 
-        
-        
-    
-    w   : 
-        
-        
-
+    ftrajs : list of trajectories of microscopic observables. Has to have
+        a one-to-one correspondence with the dtrajs. 
+    m   : Experimental average. 
+    w   : Weight of each experimental observable. Equal to 1/(2s) where s is
+        standard error of the experimental. 
     lag : int
         lag time at which transitions are counted and the transition matrix is
         estimated.
@@ -1385,6 +1379,7 @@ def estimate_augmented_markov_model(dtrajs, ftrajs, m, w, lag,
         Optional parameter with specifies the maximum number of 
         updates for Lagrange multiplier estimation.
          
+
     score_method : str, optional, default='VAMP2'
         Score to be used with MSM score function. Available scores are
         based on the variational approach for Markov processes [13]_ [14]_:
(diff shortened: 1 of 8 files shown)


franknoe:
Did you commit a vegan sausage?

pyemma/msm/api.py
```diff
@@ -1288,6 +1292,163 @@ def bayesian_hidden_markov_model(dtrajs, nstates, lag, nsamples=100, reversible=
                                   dt_traj=dt_traj, conf=conf, store_hidden=store_hidden, show_progress=show_progress)
     return bhmsm_estimator.estimate(dtrajs)
 
+def estimate_augmented_markov_model(dtrajs, ftrajs, m, w, lag,
```
franknoe: I suggest the order ``dtrajs, ftrajs, lag, m, w`` (first simulation data and parameter, then experimental constraints)

pyemma/msm/api.py
```diff
@@ -1288,6 +1292,163 @@ def bayesian_hidden_markov_model(dtrajs, nstates, lag, nsamples=100, reversible=
...
+    contains the estimated transition matrix and allows to compute a
+    large number of quantities related to Markov models.
+
+    Parameters
+    ----------
+    dtrajs : list containing ndarrays(dtype=int) or ndarray(n, dtype=int)
+        discrete trajectories, stored as integer ndarrays (arbitrary size)
+        or a single ndarray for only one trajectory.
+    ftrajs : list of trajectories of microscopic observables. Has to have
+        a one-to-one correspondence with the dtrajs.
```
franknoe: Has to have the same shape (number of trajectories and timesteps) as dtrajs

pyemma/msm/api.py
```diff
@@ -1288,6 +1292,163 @@ def bayesian_hidden_markov_model(dtrajs, nstates, lag, nsamples=100, reversible=
...
+    Parameters
+    ----------
+    dtrajs : list containing ndarrays(dtype=int) or ndarray(n, dtype=int)
+        discrete trajectories, stored as integer ndarrays (arbitrary size)
+        or a single ndarray for only one trajectory.
+    ftrajs : list of trajectories of microscopic observables. Has to have
+        a one-to-one correspondence with the dtrajs. 
+    m   : Experimental averages. 
+    w   : Weight of each experimental observable. Equal to 1/(2s) where s is
+        standard error of the experimental.
```
franknoe: If we are using Gaussians, isn't this 1/(2s^2) ?
Or how about specifying s instead? Maybe this is more intuitive.

pyemma/msm/api.py
```diff
@@ -1288,6 +1292,163 @@ def bayesian_hidden_markov_model(dtrajs, nstates, lag, nsamples=100, reversible=
...
+    m   : Experimental averages. 
+    w   : Weight of each experimental observable. Equal to 1/(2s) where s is
+        standard error of the experimental. 
+    lag : int
+        lag time at which transitions are counted and the transition matrix is
+        estimated.
+    reversible : bool, optional
+        If true compute reversible MSM, else non-reversible MSM
+    statdist : (M,) ndarray, optional
+        Stationary vector on the full state-space. Transition matrix
```
franknoe: You don't seem to have this option.

pyemma/msm/api.py
```diff
@@ -1288,6 +1292,163 @@ def bayesian_hidden_markov_model(dtrajs, nstates, lag, nsamples=100, reversible=
...
+        *  's',   'second*'
+
+    maxiter : int, optional
+        Optional parameter with specifies the maximum number of 
+        updates for Lagrange multiplier estimation.
+         
+
+    score_method : str, optional, default='VAMP2'
+        Score to be used with MSM score function. Available scores are
+        based on the variational approach for Markov processes [13]_ [14]_:
```
franknoe: These citations seem not to be in this docstring. Do they work?

franknoe: Have you thought of whether AMM and our current cross-validation approach makes sense?

pyemma/msm/tests/test_amm.py
```diff
@@ -0,0 +1,766 @@
+# This file is part of PyEMMA.
```
franknoe, line 1: Can't we reconcile this test with the test_msm or test_oom test? It seems we have a lot of redundant code here.

pyemma/msm/estimators/maximum_likelihood_msm.py
```diff
@@ -1361,3 +1361,498 @@ def OOM_sigma(self):
...
+    def _update_mhat(self): 
+        """
+          Updates mhat (expectation of observable of the Augmented Markov model) 
+        """ 
+        self.mhat = _np.dot(self.pihat.reshape((self.n_mstates_active,)), self.E_active[:]) 
+        self._update_S() 
+
+    def _update_S(self):
+        """
+          Computes slope in observable space.
```
franknoe: In this case you can write 

""" Computes slope in observable space. """

If you have text, then use

""" Title

blabla
(empty line here)
"""

pyemma/msm/estimators/maximum_likelihood_msm.py
```diff
@@ -1361,3 +1361,498 @@ def OOM_sigma(self):
...
+        self._is_estimated = True
+
+        # if active set is empty, we can't do anything.
+        if _np.size(self.active_set) == 0:
+            raise RuntimeError('Active set is empty. Cannot estimate AMM.')
+
+        # active count matrix and number of states
+        self._C_active = dtrajstats.count_matrix(subset=self.active_set)
+        self._nstates = self._C_active.shape[0]
+
```
franknoe: remove empty line (check warnings for codestyle shown on the right in PyCharm)

franknoe (changes requested):

pyemma/msm/api.py
psolsson: That's a typo. Thanks. Will fix.

pyemma/msm/api.py
psolsson: Makes sense.

pyemma/msm/api.py
psolsson: Will change.

pyemma/msm/api.py
psolsson: I'm unsure whether these actually make sense. Should I remove them for now? That was my initial thought.

pyemma/msm/tests/test_amm.py
psolsson: That could be possible now, but maybe it is good to keep them separate as a lot of the functionality does not translate directly between the models. Also to make room for some divergence in future developments. What do you think?

pyemma/msm/estimators/maximum_likelihood_msm.py
psolsson: OK.

pyemma/msm/estimators/maximum_likelihood_msm.py
psolsson: Not using PyCharm so I'm not sure what you are referring to. I'm guessing line 1683 should go?

pyemma/msm/tests/test_amm.py
psolsson: I've inherited the TestMSMDoubleWell from test_msm now and overridden special cases. This removes a lot of redundancy.

franknoe:
Cool. I think you might even be able to save most of the rest by just passing adding the AMM to the objects being tested, and only overwriting the functions starting with an "_" that actually change.

psolsson:
That should've been done with this commit.

PR #1120, opened 2017-06-14
Running covar weights input

marscher:
tica and running covar accepts lists of ndarrays as weights.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,27 +1,15 @@
 import pytest
+import sys
+
+# this is import for measuring coverage
+assert 'pyemma' not in sys.modules
 
 
 @pytest.fixture(scope='session')
 def no_progress_bars():
     """ disables progress bars during testing """
-    import subprocess
-
-    def pg_enabled():
-        out = subprocess.check_output(['python', '-c', 'from __future__ import print_function; import pyemma; print(pyemma.config.show_progress_bars)'])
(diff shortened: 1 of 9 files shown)


marscher:
I need to add a check for the weight length to match the one of input iterable.

marscher:
@fnueske Could you please comment on the fact, that zero weights are currently not handled by the formula to merge two Moments in running_moments.py https://github.com/markovmodel/PyEMMA/pull/1120/files#diff-897521a48135f9c9343041eeb18ba12fR44 ?

fnueske:
I think the updating formulas break down if the total weight of a complete chunk equals zero. Normally, it is assumed that weights are non-negative, and consequently a chunk with total weight zero can simply be discarded. In practice, especially in the Koopman re-weighting, it's a different story. I don't really know what to do here, maybe @fabian-paul has a suggestion?

marscher:
so maybe we can define an epsilon below we consider the weight absolutely zero and return the chunk with the either higher weight or just a zero vector. How are you dealing with this in Koopman re-weighting?

fabian-paul:
I'm not sure. As Feliks says, the mean of a chunk can be undefined in Koopman-reweighting.
But I think we have to assume that the mean over the whole data exists. 
So can we re-derive the summation formulas without referring to the mean of a chunk?

PR #1125, opened 2017-06-30
[running_moments] fix computation of cov_YY

marscher:
There was a bug in running moments which prevented the computation of cov_YY


fabian-paul:
Ah, here it is: https://github.com/markovmodel/PyEMMA/tree/implement_cov_YY

fabian-paul:
Ok!

PR #1136, opened 2017-07-18
Fix hmsm msm init param clone

### LINKED ISSUE #1135: Warning: 'OOMReweightedMSM' object has no attribute '_is_estimated'
There seems to be a problem with the `_is_estimated` attribute of `OOMReweightedMSM`. The following minimal example reproduces this error. Pyemma version 2.4.

```python
import pyemma
import numpy as np

dtraj = np.random.randint(0, 10, 100)
oom = pyemma.msm.estimate_markov_model(dtraj, 1, weights='oom')
hmm = oom.coarse_grain(2)
hmm.cktest()
```

marscher:
[sklearn] if an already estimated Estimator is passed, deep copy it
    
    This is mandatory for parameter studies, which clones estimators to
    provide a "fresh" and independent blue print.
    
    Eg. there was a bug in HMM.cktest(), which does a parameter study over a
    series of lag times and cloning led to forgetting derived attributes
    like the count matrix of the initial model.
    
    This is a ugly workaround for this situation and should be properly
    fixed in the future, because deep copying the object multiplies the
    memory footprint by every copy taken.

Fixes #1135 

Another potential fix for the future is the usage of the declared serialize fields in the estimator (which then of course has to be serializable, but this is mission crucial any ways). So it depends on #867 and that developers are aware that they have to declare derived/computed fields. When we use this feature, we can simply serialize the estimator and have directly everything we want (parameters and estimated quantities).

### DIFF
--- a/pyemma/_ext/sklearn/base.py
+++ b/pyemma/_ext/sklearn/base.py
@@ -70,6 +70,13 @@ def clone(estimator, safe=True):
     new_object = klass(**new_object_params)
     params_set = new_object.get_params(deep=False)
 
+    # TODO: this is a brute force method to make things work for parameter studies. #1135
+    # But this can potentially use a lot of memory in case of large input data, which is also copied then.
+    # we need a way to distinguish input parameters from derived model parameters, which is currently only ensured for
+    # estimators in the coordinates package.
+    if hasattr(estimator, '_estimated') and estimator._estimated:
+        return copy.deepcopy(estimator)
+
     # quick sanity check of the parameters of the clone
     for name in new_object_params:
         param1 = new_object_params[name]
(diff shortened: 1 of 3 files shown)


marscher:
@thempel I've added your minimal testing example to the ML-HMSM tests, so this is covered now.

In my opinion we're ready to merge, if you have nothing to add.

thempel:
Thanks, @marscher ! The bug should be resolved. Just as a side remark, I'm now getting a `PyEMMA/pyemma/msm/estimators/maximum_likelihood_msm.py:821: UserWarning: The dtrajs argument will be mandatory in the future for method hmm.` when calling the `coarse_grain()` method (same code as in #1135). Just ignore if that's discussed somewhere else...

PR #1152, opened 2017-08-29
[LaggedCovariance] added property cov_tau_tau

marscher:
We can compute the instantaneous covariance on the lagged data by setting
ctt=True, but there was no property to access it.

@fabian-paul, @franknoe Do you agree, that this property name is not optimal and do you maybe have a better suggestion?

### DIFF
--- a/pyemma/coordinates/estimation/covariance.py
+++ b/pyemma/coordinates/estimation/covariance.py
@@ -249,6 +249,11 @@ class LaggedCovariance(StreamingEstimator, ProgressReporter):
         return self._rc.cov_XY(bessel=self.bessel)
 
     @property
+    def cov_tau_tau(self):
+        self._check_estimated()
+        return self._rc.cov_YY(bessel=self.bessel)
+
+    @property
     def nsave(self):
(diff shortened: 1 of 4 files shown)


franknoe:
How about using ''C00'', ''C0t'', ''Ctt''? Then ''cov'' maps to ''C00'' and ''cov_tau'' maps to ''C0t''. For ''Ctt'' we have no other function name. Maybe we could deprecate ''cov'' and ''cov_tau'', but I'm not sure it's necessary.

marscher:
Thanks for your comment. We should deprecate cov, cov_tau first, because this Estimator has been introduced in version 2.3 (remained unchanged for three releases). The upper case properties gives the right impression, that these are matrix like objects and are synonyms of the input parameters. In sklearn it seems to be a common practice to have an underscore suffix for derived/computed/estimated properties. It also would increase readability I think.

marscher:
deprecated cov, cov_tau and added C00_, C0t_ and Ctt_ (the boolean flags to compute these are named lower case without an underscore suffix).

PR #1155, opened 2017-09-01
Models estimators serializable hdf5

marscher:
This one is going to be a killer feature :)

It allows storing a serialized model within the hdf5 file, containing the arrays. Every time one saves to the file, a new group container is created, which holds the arrays and the json string for the model is being saved into a new slot.

This way one can easily restore old models within a very short time (the models are just indexed by age). Additional attributes like human readable names and saving dates are included as well.

It is blazingly fast due to hdf5. Jsonpickle adds minimal overhead.

One problem still causes headaches... how should we disentangle np object arrays?! If I unwrap them to a list, fancy indexing will not be working after restoring the model. Also jsonpickle does not like calling flatten within a custom handler.


clonker:
Maybe you could put the output of `np.save` into a different data set in that group container with a, e.g., UUID as name which then gets referenced in the json.

marscher:
that is exactly what I am doing. All arrays which are being pickled for the save have uuids and are contained in the current group. The group is created per save, to have a point of return in case one overwrites something.

marscher:
@franknoe are you okay with introducing a HDF5 dependency for this feature?

PR #1156, opened 2017-09-01
Add new data source 'SourcesMerger'

marscher:
The class supports to combine multiple data source (eg. readers, TICA objects, etc.), which may have different dimensions. This class can then be used to provide streaming access from all underlying sources and it can be passed directly to another Estimator. The dimension axis is concatenated for this purpose.
All of the parameters for chunk, stride, skip are passed during iterator creation to ensure we get the right data.

### DIFF
--- a/pyemma/coordinates/__init__.py
+++ b/pyemma/coordinates/__init__.py
@@ -31,6 +31,7 @@ User API
    featurizer
    load
    source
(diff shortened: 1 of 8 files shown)


marscher:
"mergeable" said by @clonker

PR #1158, opened 2017-09-11
Allow to set nsamples for BMSM, BHMM and ITS

marscher:
please review @gph82

### DIFF
--- a/pyemma/_base/model.py
+++ b/pyemma/_base/model.py
@@ -146,9 +146,9 @@ class Model(object):
 class SampledModel(Model):
 
     def __init__(self, samples, conf=0.95):
+        super(SampledModel, self).__init__()
         self.set_model_params(samples=samples, conf=conf)
 
-    # TODO: maybe rename to parametrize in order to avoid confusion with set_params that has a different behavior?
     def set_model_params(self, samples=None, conf=0.95):
         self.update_model_params(samples=samples, conf=conf)
         if samples is not None:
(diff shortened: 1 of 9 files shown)


marscher:
not sure what happened to the Appveyor tests, it seems like there is another run_test script on PATH which shadows the script we actually wanted to execute...

marscher:
Since sampling in all three classes is a post-estimation thing, it still complies with the quoted sklearn patterns. Both MSM classes performed the sampling in the _estimate method, but for this step only "real" estimated quantities are used, so splitting the sampling from the actual estimation (code wise) increases both readability and functionality. Why would a convenience function avoid confusion or potential bugs (this function would have to "fiddle" around with internal estimator properties and people would expect such a functionality within the object)?

The PR is a bit messy, because I also fixed some other issues along the way.

Just a question about BHMM, why are there two discrete_hmm estimations? One by calling super()._estimate or use the hmm_init parameter and another for sampling input - I  could not spot a difference in the results produced.

franknoe:
I don't understand. In the Bayesian classes, sampling is an integral part of the estimation, so estimation is not completed before sampling is.
As for this PR, maybe we can separate fixes needed anyway from this sampling issue, to make it more understandable. 


marscher:
On 11.09.2017 19:52, Frank Noe wrote:
> 
>      
> I don't understand. In the Bayesian classes, sampling is an integral part of the estimation, so estimation is not completed before sampling is.
> As for this PR, maybe we can separate fixes needed anyway from this sampling issue, to make it more understandable.
I didn't say estimation is complete prior sampling. I'm just saying that 
there are two steps, generate a matrix to be sample and the sampling it 
self.

PR #1159, opened 2017-09-12
[MLMSM] use pi from msmtools tmatrix estimation

marscher:
For reversible estimates we get pi directly from msmtools.estimation.tmatrix. So we use it for the cases where it applies. The tests were checking eigenvalues within a numerical tolerance of 1e-12, so we set max_err accordingly.

Done with the aid of @fabian-paul

### DIFF
--- a/pyemma/msm/estimators/maximum_likelihood_msm.py
+++ b/pyemma/msm/estimators/maximum_likelihood_msm.py
@@ -1117,15 +1117,6 @@ class MaximumLikelihoodMSM(_MSMEstimator):
         if isinstance(P, tuple):
             P, statdist_active = P
 
-        # continue sparse or dense?
-        if not self.sparse:
-            # converting count matrices to arrays. As a result the
-            # transition matrix and all subsequent properties will be
-            # computed using dense arrays and dense matrix algebra.
-            self._C_full = self._C_full.toarray()
-            self._C_active = self._C_active.toarray()
-            P = P.toarray()
-
         # Done. We set our own model parameters, so this estimator is
         # equal to the estimated model.
         self._dtrajs_full = dtrajs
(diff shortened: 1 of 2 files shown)


marscher:
@franknoe, @fabian-paul would it hurt anybody, if we increase the default max_err value to 1e-12?

marscher:
https://travis-ci.org/markovmodel/PyEMMA/jobs/274658288 shows TPT pathway decomposition fails, because of using pi from the iteration formula instead of the eigendecomposition...

marscher:
Since we have repaired sparse/dense estimation previously, the reactive flux decomposition now works as well. I guess there were some kind of inconsistencies if we estimated dense and then use sparse routines to decompose.

marscher:
The estimation method returns an invalid pi for a symmetric estimate. Eg. even the stationary relation is violated (P.mu != mu).

If I add the following to the MLMSM._estimate method, the first numpy testing assertion already fails. I think this is something serious.
``` python
        # msmtools returns a tuple for statdist_active = None.
        if isinstance(P, tuple):
            P, statdist_active = P
            _np.testing.assert_allclose(P.dot(statdist_active), statdist_active)
            if self.reversible:
                _np.testing.assert_allclose(statdist_active.dot(P), P.T.dot(statdist_active))
```

yields:
```
>           _np.testing.assert_allclose(P.dot(statdist_active), statdist_active, atol=atol)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=1e-12
E           
E           (mismatch 100.0%)
E            x: array([ 0.250438,  0.249998,  0.250557,  0.250871])
E            y: array([ 0.239957,  0.266248,  0.252854,  0.24094 ])
```
However the estimation routine returned no error code and reported convergence.

marscher:
Either we fix this issue in msmtools or we rely on diagonalizing the transition matrix for computing the stationary measure. What do you think?

franknoe:
Huh, why does this happen now? Shouldn't the test have failed before the 2.5 release. 
We should go back to the last version that worked and retrace it from there. Or does this only happen with new numpy?


marscher:
I computed the stationary measure wrong, sorry for spreading fake news!

marscher:
This PR depends on markovmodel/msmtools#107

PR #1178, opened 2017-11-06
Msm score cv not in place

marscher:


marscher:
the windows tests failed unrelated, but indicate a problem with minRMSD and multi-threading in the assignment routine...

marscher:
@gph82 Do you agree, that this solution seems to be the correct way of fixing this?

gph82:
not familiar with "clone", but it looks like that's the way it should be. MSMs changing after estimating is very confusing and unpredictable, for e.g. for functions like MSM.sample_by_state or MSM.sample_by_distributiion. I know @franknoe is busy RN, otherwise LGTM

marscher:
clone() just clones the Estimators parameters (not the estimated model,
quantities). Since we do a re-estimation here, the latter should not
inflict any harm on the method itself.

PR #1180, opened 2017-11-06
replaced cython extension for OrderedSet with pythonic (3) solution

marscher:
Reduces code complexity (therefore increasing maintainability) at the expense of a little more memory overhead by using OrderedDict as storage.


marscher:
memory footprint is very large. Closing.

PR #1184, opened 2017-11-14
[featurizer] added alignment to reference for selection based features.

marscher:


marscher:
@gph82 can you take a look please?

pyemma/coordinates/data/featurization/misc.py
```diff
@@ -219,3 +219,24 @@ def __hash__(self):
...
+        self.prefix_label = 'aligned ATOM:'
+
+    def __hash__(self):
+        h = super(AlignFeature, self).__hash__()
+        h ^= hash(self.ref)
+        return h
+
+    def transform(self, traj):
+        aligned = traj.superpose(reference=self.ref, atom_indices=self.atom_indices,
+                                 ref_atom_indices=self.ref_atom_indices)
```
gph82: I would add a warning in the docs that using the alignment modifies the input md.Trajectory object (because traj.superpose returns self).

gph82:
Otherwise, LGTM

marscher:
I've thought about adding a 'in_place' option, but actually the
Trajectory object is never seen by the user, so it should be safe to
modify it in place.

gph82:
except when using the transform method on its own!

marscher:
On 11/14/2017 04:39 PM, Guillermo Pérez-Hernández wrote:
> except when using the transform method on its own!

Are you doing that? It is not the usual case.

gph82:
I am doing that very often!

marscher:
ok, added this just for you :) but it is not bound to the featurizer api, but lives directly in the feature class.

gph82:
sure, that is fine. You can make it even just in the doc of the .transform method. But I definitely use feat.transform() very often, quite a lot indeed, and it is actually very useful. I don't think is is unusual

gph82:
oh, I just meant add to the doc that there's an inplace transformation, not the option....

gph82:
I know @cwehmeyer will appreciate this, I wrote a workaround using add_custom_func

PR #1188, opened 2017-11-23
Kmeans cpp loop

marscher:
added method to cpp extension to keep running in the extension as long as possible.

### DIFF
--- a/pyemma/coordinates/clustering/include/bits/kmeans_bits.h
+++ b/pyemma/coordinates/clustering/include/bits/kmeans_bits.h
@@ -143,6 +143,36 @@ KMeans<dtype>::cluster(const np_array &np_chunk, const np_array &np_centers, int
 }
 
 template<typename dtype>
+std::tuple<py::array_t<dtype>, int, int> KMeans<dtype>::cluster_loop(const np_array& np_chunk, np_array& np_centers,
+                                                      int n_threads, int max_iter, float tolerance) const {
+    int it = 0;
+    bool converged = false;
+    dtype rel_change = 0;
+    dtype prev_cost = std::numeric_limits<dtype>::max();
+
+    do {
+        np_centers = cluster(np_chunk, np_centers, n_threads);
+        auto cost = costFunction(np_chunk, np_centers, n_threads);
+        prev_cost = cost;
+        if(cost != 0) {
+            rel_change = std::abs(cost - prev_cost) / cost;
+        }
+        if(rel_change <= tolerance) {
+            converged = true;
+            break;
+        } else {
+            if(! callback.is_none()) {
+                callback();
+            }
+        }
+
+        it += 1;
+    } while(it < max_iter);
+
+    return std::make_tuple(np_centers, converged, it);
+}
+
+template<typename dtype>
 dtype KMeans<dtype>::costFunction(const np_array &np_data, const np_array &np_centers, int n_threads) const {
     auto data = np_data.template unchecked<2>();
     auto centers = np_centers.template unchecked<2>();
--- a/pyemma/coordinates/clustering/include/bits/kmeans_bits.h
+++ b/pyemma/coordinates/clustering/include/bits/kmeans_bits.h
@@ -142,34 +142,34 @@ KMeans<dtype>::cluster(const np_array &np_chunk, const np_array &np_centers, int
     return return_new_centers;
 }
 
+
 template<typename dtype>
 std::tuple<py::array_t<dtype>, int, int> KMeans<dtype>::cluster_loop(const np_array& np_chunk, np_array& np_centers,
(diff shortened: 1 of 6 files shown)


pyemma/coordinates/clustering/include/bits/kmeans_bits.h
```diff
@@ -142,6 +142,36 @@ KMeans<dtype>::cluster(const np_array &np_chunk, const np_array &np_centers, int
     return return_new_centers;
 }
 
+
+template<typename dtype>
+std::tuple<py::array_t<dtype>, int, int> KMeans<dtype>::cluster_loop(const np_array& np_chunk, np_array& np_centers,
```
clonker: you could typedef this `std::tuple<py::array_t<dtype>, int, int>` so it becomes more descriptive

pyemma/coordinates/clustering/include/bits/kmeans_bits.h
```diff
@@ -142,6 +142,36 @@ KMeans<dtype>::cluster(const np_array &np_chunk, const np_array &np_centers, int
...
+    dtype rel_change = std::numeric_limits<dtype>::max();
+    dtype prev_cost = 0;
+    do {
+        np_centers = cluster(np_chunk, np_centers, n_threads);
+        auto cost = costFunction(np_chunk, np_centers, n_threads);
+        rel_change = (cost != 0.0) ? std::abs(cost - prev_cost) / cost : 0;
+        prev_cost = cost;
+        if(rel_change <= tolerance) {
+            converged = true;
+            break;
```
clonker: I'd write this as `while(it < max_iter && !converged)` instead of using break

pyemma/coordinates/clustering/include/bits/kmeans_bits.h
```diff
@@ -142,6 +142,36 @@ KMeans<dtype>::cluster(const np_array &np_chunk, const np_array &np_centers, int
...
+                /* Acquire GIL before calling Python code */
+                py::gil_scoped_acquire acquire;
+                callback();
+            }
+        }
+
+        it += 1;
+    } while(it < max_iter);
+    int res = converged ? 0 : 1;
+    return std::make_tuple(np_centers, res, it);
```
clonker: perhaps it makes sense to `std::move(...)` the arguments into `std::make_tuple`

pyemma/coordinates/clustering/include/bits/kmeans_bits.h
```diff
@@ -168,7 +198,7 @@ dtype KMeans<dtype>::costFunction(const np_array &np_data, const np_array &np_ce
 template<typename dtype>
 typename KMeans<dtype>::np_array KMeans<dtype>::
 initCentersKMpp(const np_array &np_data, unsigned int random_seed, int n_threads) const {
-
+    py::gil_scoped_release release;
```
clonker: 👍

pyemma/coordinates/clustering/include/kmeans.h
```diff
@@ -33,6 +33,9 @@ class KMeans : public ClusteringBase<dtype> {
      */
     np_array cluster(const np_array & /*np_chunk*/, const np_array & /*np_centers*/, int /*n_threads*/) const;
 
+    std::tuple<py::array_t<dtype>, int, int> cluster_loop(const np_array & /*np_chunk*/, np_array & /*np_centers*/,
```
clonker: typedef remark from above applies here as well

pyemma/coordinates/clustering/kmeans.py
```diff
@@ -226,23 +226,14 @@ def _estimate(self, iterable, **kw):
...
-                    self._converged = True
-                    self._logger.info("Cluster centers converged after %i steps.", it + 1)
-                    break
-                else:
-                    self._progress_update(1, stage=1)
-                it += 1
-            if not self._converged:
+        with self._progress_context(stage=1):
+            if self.show_progress:
+                self._inst.set_callback(lambda: self._progress_update(1, stage=1))
```
clonker: how about registering the callback as an argument to `cluster_loop`, keeping it localized? I very much like this way of registering the progress update btw

clonker (approved):
Some very pedantic remarks that mostly boil down to a matter of taste in code style, I guess. Looks good to me!

marscher:
thanks for the review @clonker, now I remember why I set the callback to be a member of the class. One can not pass None if the type of the argument is py::function...

clonker:
You can, however, use `py::object`:
```cpp
void fun(const py::object& callback = py::none()) {
    if(callback.is_none()) { /* do this */ }
    else { /* do that */ }
}
callback(args);
```

marscher:
thanks, this was the same approach for the ctor :)

Issue #1195, opened 2017-12-03
git install not working


RobertArbon:
I did: 

```
Research ➤ git clone git@github.com:RobertArbon/PyEMMA.git
cd PyEMMA
PyEMMA ➤ python setup.py develop                                                                                                                               git:7ed62207
setup.py:274: UserWarning: using git, require cython
  warnings.warn('using git, require cython')
Submodule 'pybind11' (https://github.com/pybind/pybind11) registered for path 'pybind11'
Cloning into '/Users/robert_arbon/Google Drive/Research/PyEMMA/pybind11'...
Submodule path 'pybind11': checked out '1df91d36cfa997c8987889234304ea5e37cc3e68'
```
and then I get the following error: 
```
pybind11/include/pybind11/common.h:126:10: fatal error: 'forward_list' file not found
#include <forward_list>
         ^~~~~~~~~~~~~~
1 error generated.
error: command '/usr/bin/clang' failed with exit status 1
```

This is strange because I did a git install just a few days ago. 

I don't know enough about submodules or pybind11 to comment much further, except to say that:

- The problem seems to be that the submodule is updating to an old version of `pybind11`  - version 2.1 and that requires the header file `forward_list` which can't be found. 

- I notice that the latest version of pybind11 is 2.2 and that has deprecated the `common.h` file. 

I've tried specifying the stable branch in `.gitmodules` file i.e. 
```
[submodule "pybind11"]
	path = pybind11
	url = https://github.com/pybind/pybind11
        branch = stable
```
but the command line call in `setup.py` seems to override that. 

I've also checked out commit `7ed622076b004f8a114f7f5b42fe9585a566e8e1` on `devel` from 3 days ago but the problem persists.  

I've tried manually updating the PyEMMA/pybind11/ submodule to the latest version before running `setup.py` but the `        cmd = "git submodule update --init {mod}"` seems to override that and install the old version (2.1). 

I'm on OSX Sierra.

marscher:
Thanks the report. Ain't clang cxx11 compatible? Any how I pushed a branch to use the latest release of pybind. What is the exact clang invocation here?

marscher:
I guess just updating the version will not solve your problem, because the compiler flags should have told clang to use at least cxx11 standard to compile the code. Probably apple broke the gcc command line options compatibility again...

RobertArbon:
[...]

PR #1197, opened 2017-12-04
Fix compilation darwin

marscher:
use flag setting pattern of pybind python example. Need to distinguish between c and c++ extensions.


marscher:
Fixes #1195

Issue #1189, opened 2017-11-24
BayesianHMSM fails for option init_hmsm


marscher:
the estimate method copies over various attributes, but does not catch all attributes of importance and just crashes. @thempel you experienced this problem as well and you got a workaround for it, right?

Note that this option was simply not tested (and therefore broken). So it is deleted in the current devel branch. We can restore it as soon we got a working impl and test.

thempel:
Yes, a functioning temporary workaround is to simply update `BayesianHMSM.__dict__` in the appropriate case of the `BayesianHMSM._estimate()` method.

```python
exclude_keys = ['stride', '_estimated']
self.__dict__.update(
{k: self.init_hmsm.__dict__[k] for k in self.init_hmsm.__dict__.keys() if k not in exclude_keys})

```
It would be better to whitelist, though, if we use this approach at all. Also one should discuss which arguments should be used in case a user delivers different choices for the init_hmm and the Bayesian one.

marscher:
fixed by #1199

PR #1199, opened 2017-12-07
fix for Initialized bhmm

thempel:
Fixes the Bayesian HMM estimation initialized with a ML HMM #1189.

- is the `self.observe_nonempty` handled correctly?
- are there any wrong implications that I don't immediately see by copying the attributes listed in the `copy_attributes` list?
- should we check for more user choices?


marscher:
what about the interplay of lag and stride?

thempel:
Right, I've added support for a stride that is different to the one of the init_hmsm (and a check that we still have the same observable set).

marscher:
thanks a lot!

PR #1205, opened 2017-12-18
Serializable review

marscher:
@clonker I hope I included all of your requested changes.

### DIFF
--- a/pyemma/__init__.py
+++ b/pyemma/__init__.py
@@ -55,8 +55,8 @@ def load(filename, model_name='latest'):
         the instance containing the same parameters as the saved model/estimator.
 
     """
-    from ._base.serialization.serialization import load as _load
-    return _load(filename, model_name)
+    from ._base.serialization.serialization import SerializableMixIn
+    return SerializableMixIn.load(file_name=filename, model_name=model_name)
 
 
 def _version_check(current, testing=False):
--- a/pyemma/_base/estimator.py
+++ b/pyemma/_base/estimator.py
@@ -359,24 +359,14 @@ def estimate_param_scan(estimator, X, param_sets, evaluate=None, evaluate_args=N
         return res
 
 
+# we do not want to derive from Serializable here, because this would make all children serializable.
+# However we guide serializable children, what to store/restore.
 class Estimator(_BaseEstimator, Loggable):
     """ Base class for pyEMMA estimators
 
     """
     # flag indicating if estimator's estimate method has been called
     _estimated = False
-    _serialize_version = 0
-    _serialize_fields = ('_estimated', 'model')
-
-    def __new__(cls, *args, **kwargs):
-        if hasattr(cls, '_serialize_fields'):
-            new_fields = cls._get_param_names()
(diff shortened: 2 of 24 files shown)


clonker:
Ive just scanned through:
  - you still use `__repr__` instead of `__str__` as a human readable output of model
  - there are a couple places where you use, e.g., `saved_streaming_chain` as a standalone string instead of having it centralized somewhere - but i guess this is just a matter of taste in the end
  - im not sure if you are raising if a group in a h5 file already exists (same model name, same file)
  - you dont raise in case of a model that was saved with an older version (`if klass_version > klass._serialize_version`)
  - minRMSD feature with reference frame changes?
  - custom feature serialization? perhaps use dill instead of pickle at that point?

Other than that: 
  - Nice job on the fluent api for `Modifications`, i would just declase the `ops` attribute private
  - rest (except for above remarks) looks good to me

marscher:
Thanks for scanning :)

On 12/18/2017 02:22 PM, Moritz Hoffmann wrote:
> Ive just scanned through:
> 
>   * you still use |__repr__| instead of |__str__| as a human readable
>     output of model
>   * there are a couple places where you use, e.g.,
>     |saved_streaming_chain| as a standalone string instead of having it
>     centralized somewhere - but i guess this is just a matter of taste
>     in the end
>   * im not sure if you are raising if a group in a h5 file already
>     exists (same model name, same file)
>   * you dont raise in case of a model that was saved with an older
>     version (|if klass_version > klass._serialize_version|)
>   * minRMSD feature with reference frame changes?
>   * custom feature serialization? perhaps use dill instead of pickle at
>     that point?
The problem with pickling arbitrary function is that we can not know if
the function is pure, eg. it does not depend on globals form the outer
scope. We would then also need to pickle these potentially creating a
very huge dependency. This holds true especially for jupyter notebook
users, which have "everything" in the global namespace.

So I would suggest to just trigger a warning, when serializing custom
features, that we refuse to store the function.

I think this is the only solution, which does not cause serious trouble.
Or we require the input function to be pure - which is somehow
impossible in Python, because everything can be modified.

I mean it is implemented in dill, even with a "recursive" option to drag
in the functions dependencies, but this can be a super long rat tail
(even modules etc. can be pickled), for which it would be hard or nearly
impossible to maintain "future" compatibility with.

> 
> Other than that:
> 
>   * Nice job on the fluent api for |Modifications|, i would just declase
>     the |ops| attribute private
>   * rest (except for above remarks) looks good to me

clonker:
Youre welcome!

True i hadn't thought about the trailing dependencies. Then raising a warning and just storing a dummy function seems to be best (instead of failing with a nasty pickling error).

marscher:
repr vs. str inherited from scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/de0581aea7fadae6646d86d92938e59e6ddfa043/sklearn/base.py#L273

Probably one should report this to them 👍

marscher:
@clonker made sure to include all of your change requests this time :)

clonker (approved):
looks good to me 👍

PR #1214, opened 2018-01-05
Thermo: fix subtle bug in detection of the unbiased state

### LINKED ISSUE #1204: thermo.estimate_multi_temperature() return object has no attribute 'msm'
pyemma: 2.4 / np111py27_1 from conda-forge
RedHat Linux Version: Linux version 2.6.32-642.6.2.el6.x86_64 (mockbuild@x86-031.build.eng.bos.redhat.com) (gcc version 4.4.7 20120313 (Red Hat 4.4.7-17) (GCC) ) #1 SMP Mon Oct 24 10:22:33 EDT 2016

When using thermo.estimate_multi_temperature(), the returned object is of type TRAM and not MEMM or MultiThermModel, as the documentation here suggests: http://emma-project.org/latest/api/generated/thermo-api/pyemma.thermo.estimate_multi_temperature.html#pyemma.thermo.estimate_multi_temperature

I then run into the issue that the returned object has its attribute `msm` = `None`, which the tutorial (http://emma-project.org/latest/generated/PyEMMA.thermo.estimate_multi_temperatur_-_asymmetric_double_well.html) suggests should be not `None` with this returned object. Also, I am not sure if this is an issue, but the object's `TRAM.models[0].stationary_distribution` is not the same as the `TRAM.stationary_distribution`. I think I am missing something so please do let me know what to do differently to get the resultant MSM from the TRAM estimation.

An example in a tar-ball is attached, the script run_test.py will print out the difference between the standard deviation and the type of the objects that were produced, as well as return the output of `thermo.estimate_multi_temperature()` if you want to test it further. It also has a list of my conda list in "conda_list.txt" in the main directory. 

[debug_send.tar.gz](https://github.com/markovmodel/PyEMMA/files/1566203/debug_send.tar.gz)

fabian-paul:
fixes #1204

### DIFF
--- a/pyemma/thermo/util/util.py
+++ b/pyemma/thermo/util/util.py
@@ -322,7 +322,7 @@ def _get_multi_temperature_bias_sequences(
                     * energy_traj[:, _np.newaxis])
     else:
         # non-reduced case and temperatures given
-        kT = temperatures
+        kT = temperatures.copy()
         rT = reference_temperature
         if temp_unit.lower() == 'c':
             kT += conversion_shift_Celsius_to_Kelvin


fabian-paul:
BTW: in a future version we shouldn't allow the `temp_trajs` and `reference_temperature` parameters of `estimate_multi_temperature` to have different `dtype`s. This can lead to even more subtly bugs.

PR #1217, opened 2018-01-08
[serialization] made thermo estimators serializable.

marscher:


marscher:
@cwehmeyer When you have time, please have a look at the base class ThermoBase and if you are fine with with that change. It just adds some commonly shared attributes among the estimators. If an attribute is not defined, the "Undefined" type is returned, eg. for 'temperatures' in umbrella sampling data.

marscher:
@cwehmeyer this is the new base class for thermo estimators: https://github.com/markovmodel/PyEMMA/pull/1217/files#diff-05d670e42c6acddfbc5968ac7084b366

marscher:
@giopina Feel free to test with your tram notebook, when you have time. Thanks a lot.

PR #1218, opened 2018-01-10
[featurizer] added support for all sidechain dihedrals

### LINKED ISSUE #456: Featurizer: support all flexible torsions
Currently we have phi/psi and chi1 torsions. Why only chi1? Can we do something like add all sidechain torsions?

marscher:
Note that dihedrals are not packed in groups, since not all residues
contain the same amount of CHI dihedrals.

fixes #456

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,6 +35,6 @@ on_finish:
   # upload test results to AppVeyor
  - ps: |
      $wc = New-Object 'System.Net.WebClient';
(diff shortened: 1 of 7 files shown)


marscher:
Note that the 2 tests which includes chi5 are currently failing because of a bug in mdtraj.

marscher:
https://github.com/mdtraj/mdtraj/pull/1322

marscher:
as discussed with @sstolzenberg, we deprecate add_chi1_torsion in favour of add_sidechain_torsions

PR #1223, opened 2018-01-16
Fixes 

marscher:
* Fixed setting of description for notebook progress bars.
* fix multiple lag time estimation in thermo api
* fixed mapping from full state in thermo notebooks


marscher:
The new mapping had been introduced in #902 a long time ago. The notebooks now reflect this change. The metastable sets are now correctly separated in the multi_temp notebook.

PR #1231, opened 2018-01-25
[its] robust dtraj hashing; save missing field

marscher:


marscher:
pinned conda-build to 3.2 because of https://github.com/conda/conda-build/issues/2651

PR #1233, opened 2018-01-29
added miniconda and pyemma install script [ci skip]

marscher:
Usage purpose is one-shot installation eg. for workshops, where people don't have miniconda installed. The installer first checks, if conda is available. If thats the case it creates an environment named pyemma containing the packages. If not, it downloads miniconda, installs it and create the env afterwards.
The bashrc is updated by default to include the new miniconda installation.

The target directories and env name can be set via parameters to the script. In addition some things can be opted out, like bashrc modification and optional packages (currently it installs notebook too).

Tested on Linux and OSX.

### DIFF
--- a/install_miniconda+pyemma.sh
+++ b/install_miniconda+pyemma.sh
@@ -96,16 +96,18 @@ function install_pyemma {
     packages="pyemma numpy"
     if [ ! minimal ]; then
         # add optional packages.
-        packages="$packages notebook"
+        packages="$packages notebook ipython"
     fi
     # we add numpy here to get the openblas variant (instead of MKL).
     echo "creating pyemma environment..."
     conda=${target}/bin/conda
-    # &> /dev/null
     if [[ -z $($conda env list | grep ${env_name}) ]]; then
         echo "env '$env_name' does not exist. Create new one."
         $conda create -n ${env_name} ${packages} -y -c conda-forge
+	source activate ${env_name}
         $conda config --env --add channels conda-forge
+	echo "execute the following command to activate your newly created pyemma environment:"
+	echo "> source activate ${env_name}"
     else
         echo "environment with name='${env_name}' already exists. Please specify a different name by -e new_name"
         exit 1
@@ -113,7 +115,7 @@ function install_pyemma {
 }
 
 # probe for conda, install miniconda otherwise.
-if hash conda 2>/dev/null && [ $ignore -eq 0 ]; then
+if [ -x "$(command -v conda)" ] && [ $ignore -eq 1 ]; then
     echo "Found conda. Will install to a new environment named ${env_name}"
     install_pyemma
 else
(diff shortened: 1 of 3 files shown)


install_miniconda+pyemma.sh
```diff
@@ -0,0 +1,127 @@
...
+        packages="$packages notebook"
+    fi
+    # we add numpy here to get the openblas variant (instead of MKL).
+    echo "creating pyemma environment..."
+    conda=${target}/bin/conda
+    # &> /dev/null
+    if [[ -z $($conda env list | grep ${env_name}) ]]; then
+        echo "env '$env_name' does not exist. Create new one."
+        $conda create -n ${env_name} ${packages} -y -c conda-forge
+        $conda config --env --add channels conda-forge
```
clonker: I am not a 100% sure about this one but i think it alters the `.condarc` in $HOME, it might be nice to create a `.condarc` in the newly created environment instead so everything is encapsulated

install_miniconda+pyemma.sh
```diff
@@ -0,0 +1,127 @@
...
+       e) env_name=$OPTARG;;
+       i) ignore=0;;
+       h) usage
+         exit 0;;
+       *) echo "unknown option/combination"; exit 0;;
+   esac
+done
+
+
+function add_to_bashrc {
```
clonker, line 38: Doesn't the miniconda installer do this on its own anyway?

install_miniconda+pyemma.sh
```diff
@@ -0,0 +1,127 @@
...
+        $conda create -n ${env_name} ${packages} -y -c conda-forge
+        $conda config --env --add channels conda-forge
+    else
+        echo "environment with name='${env_name}' already exists. Please specify a different name by -e new_name"
+        exit 1
+    fi
+}
+
+# probe for conda, install miniconda otherwise.
+if hash conda 2>/dev/null && [ $ignore -eq 0 ]; then
```
clonker: This did not work on my mac, this however did:
```bash
if [ -x "$(command -v conda)" ] && [ $ignore -eq 1 ]; then
```

clonker (changes requested):
Nice script! However it did not work as expected on my mac where a conda installation was already existing (i suspect the same issue on linux) and 
I would probably prefer to have ipython installed right away, otherwise it might lead to confusion:
Ipython already being present in an existing conda installation and sourcing the new environment + launching ipython will give you the root environment...

install_miniconda+pyemma.sh
marscher: only if the installer is invoked interactively. We use batch mode and so it does not modify bashrc.

install_miniconda+pyemma.sh
marscher: Actually I thought conda config --env will place this stuff in the env. It seems like its choosing the active env.

install_miniconda+pyemma.sh
marscher: Will change that. Thanks!

install_miniconda+pyemma.sh
clonker: Which is the root env in this case, right?

marscher:
Thank you very much for the quick review! I've added your requsted changes.

clonker (approved):

clonker:
Looks good to me!

PR #1234, opened 2018-02-01
Python-2.7 deprecation and compatibility.

marscher:
The serialization code is for now Python3 only, because it has been written with assumptions in mind which are only valid for Py3. If one calls the (public) save method for serializable things under Python2 one will get an exception (with another upgrade notice).
However the pickup of serializable fields still works, so in case of pickling only desired stuff will be serialized.

Also fixed the doctests Numpy formatting, by passing the new legacy option in numpy.print_options.

Py27 is re-added to the build matrices of Travis and Appveyor.

The deprecation notice was on devel for 5 months, but never released. This is why we are doing this now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ env:
     - PYEMMA_NJOBS=1
     - MACOSX_DEPLOYMENT_TARGET=10.9
   matrix:
+    - CONDA_PY=2.7
     - CONDA_PY=3.5
     - CONDA_PY=3.6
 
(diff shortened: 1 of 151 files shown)


marscher:
tests will fail until the new build of bhmm is there (merged recently on conda-forge). https://github.com/conda-forge/bhmm-feedstock/pull/5

marscher:
Under windows we can *not* support Python2.7 because it requires the usage of a super outdated compiler which lacks support for modern C++. This renders most of our extensions unusable. I don't think this is a huge draw-back, because we intend to keep backward compat only for one further release.

For technical details and discussion have a look here: https://github.com/conda-forge/conda-forge.github.io/issues/64

PR #1237, opened 2018-02-05
VAMP estimator and cktest

marscher:


marscher:
I removed those merge commits which deleted absolute imports (important for py27 backward compat). Now the PR is is much more readable. Any volunteers to review this?

PR #1242, opened 2018-02-06
Bugfixes and export_to_hdf5

### LINKED ISSUE #1241: deprecate 'write_to_csv' and introduce hdf5 exporter
both methods just pull all transformed data through an iterator and dump it to a file. For efficiency we should use HDF5

### LINKED ISSUE #1243: DataSourceIterator uses _itraj in trajectory_lengths
The documentation of _itraj says, this property shall not be used within iterator loops, because it returns the value of the upcoming trajectory and not the actual one. However this method is public.
So whenever one calls trajectory_lengths on a DataSourceIterator (eg. on a StreamingTransformerIterator) one will get the lengths of the upcoming trajectory, which is kind of confusing and might lead to subtle bugs.

This method is used within the iterator implementations of:
* FeatureReader
* CSVReader
* SourcesMerger

If the tests for these readers/iterators only contain equal lengths trajectories, we would not catch this bug. I will investigate this further.

marscher:
* use clang on OSX and gcc on Linux.
* fixes #1243 
* fixes #1241 
* fix dimension in VAMPModel (does not depend on call of diagonalize)
* fix kinetic map scaling in VAMPModel
* RunningCovar is now serializable
* partial_fits of VAMP and RunningCovar are now serializable 
* some refactoring


marscher:
@fabian-paul https://ci.appveyor.com/project/marscher/pyemma/build/1.0.3387/job/rib38lenmo12e6io#L1449 I guess we would need another tolerance exception for MKL on 32 bit windoze?

marscher:
pytest durations argument is the culprit for this collector issues...

PR #1246, opened 2018-02-12
[vamp] added missing ctor to VAMPModel to initialize all fields properly

marscher:
Somehow this case was not covered.
I got an exception during the diagonalization, because self._scaling does not exist. This was caused by a missing constructor/invocation of the setter. I think update_model_params does not play nicely with properties, because it has been written with the assumption that attributes are always plain. Eg. it does only update/set the value, if it does not exist or the value is not None.
Perhaps we should change that too.


marscher:
Note this bug only happend if the Model is directly instantiated, so the VAMP estimator and its projection was unaffected.

PR #1247, opened 2018-02-13
[Iterable] fix default_chunksize calculation.

marscher:
I think we need to release 2.5.1...


marscher:
approved offline by @clonker

PR #1252, opened 2018-02-17
Disable random access optimization in patches

clonker:
In case of large strides or large chunks in terms of memory consumption an optimization is activated in `patches.py` which leads to taking the random-access mode rather than chunked reading. I observed that this can lead to undesired output in case of the lagged iterator. This is why i propose to deactivate this optimization for now.

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-2.5.1 (02-14-2018)
+2.5.1 (02-17-2018)
 ------------------
 
 Quick fix release to repair chunking in the coordinates package.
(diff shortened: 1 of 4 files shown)


marscher:
I think it is finally time to try to fix this memory issue in mdtraj. Seeking the desired frames should always be more io/memory efficient than reading the whole file and discarding frames afterwards.
Thank you very much for this detective like work! You're being promoted to chief inspector 🕵️ 

The new test is catching any faulty behaviour from the feature reader iterator, right?

I think this opt was introduced to work around this:
https://github.com/markovmodel/PyEMMA/issues/480

We should provide a mdtraj patch for xtc soon, then we can delete this optimization code.
https://github.com/mdtraj/mdtraj/blob/fe45f86d44f77f333424a2ba458fa14e3706e6d1/mdtraj/formats/xtc/xtc.pyx#L394

clonker:
The tests cover the cases where the io optimized iterator is chosen and where it is not. So if one sets
```python
iterload._DEACTIVATE_RANDOM_ACCESS_OPTIMIZATION = False
```
one of them fails. So yes, I think they should catch any faulty behavior from the feature reader (without any selected features).

marscher:
That is great. Thank you so much!

PR #1255, opened 2018-02-19
[CKTest] copy test model to avoid side effects.

### LINKED ISSUE #1254: CKtest broken?
It seems that doing the CK-test on the same MSM object twice (or more often) breaks something with the new estimates, i.e. the plot_cktest() function shows constant probabilities. Pyemma release 2.5.1. Minimal example:

```python
import pyemma
import pyemma.datasets
M = pyemma.msm.estimate_markov_model(pyemma.datasets.load_2well_discrete().dtraj_T100K_dt10, lag=2)
ck = M.cktest(2)
ck = M.cktest(2)
pyemma.plots.plot_cktest(ck);
```

thempel:
Somehow the propagation of the test models
transition matrix induces side effects on the model.

Fixes #1254

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 =========
 
+2.5.2 (??-??-????)
+------------------
+
+**Fixes**:
+
+- msm: Chapman Kolmogorov validator ensures there are no side effects on the tested model. #1255
+
+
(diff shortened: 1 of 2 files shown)


thempel:
@marscher thanks a lot!!

marscher:
I think we should investigate, which side-effect in MSM actually caused this, since it might be an issue in other spots too.

PR #1261, opened 2018-02-25
[coordinates.transform] Sparse sampling TICA transformer

florianlitzinger:
This introduces the NystroemTICA transformer that implements a sparse sampling variant of TICA based on a selection of column subsets of the relevant covariance matrices. To that end, additional arguments to coordinates.estimation.covariance, _ext.variational.estimators.moments, and _ext.variational.estimators.running_moments have been introduced that allow for the restriction of the covariance computation to the diagonal or a subset of columns, respectively. The tests have been extended to cover the new features.

### DIFF
--- a/pyemma/_ext/variational/estimators/moments.py
+++ b/pyemma/_ext/variational/estimators/moments.py
@@ -509,8 +509,8 @@ def _M2_sparse_sym(Xvar, mask_X, Yvar, mask_Y, weights=None, column_selection=No
     else:
         mask_Xk = mask_X[column_selection]
         mask_Yk = mask_Y[column_selection]
-        Xvark = Xvar[:, _filter_variable_indices(mask_X, column_selection)].copy()
-        Yvark = Yvar[:, _filter_variable_indices(mask_Y, column_selection)].copy()
+        Xvark = Xvar[:, _filter_variable_indices(mask_X, column_selection)]
+        Yvark = Yvar[:, _filter_variable_indices(mask_Y, column_selection)]
 
     Cxxyy = np.zeros((len(mask_X), len(mask_Yk)))
     Cxxyy[np.ix_(mask_X, mask_Xk)] = _M2_dense(Xvar, Xvark, weights=weights)
@@ -551,8 +551,8 @@ def _M2_symmetric(Xvar, Yvar, mask_X=None, mask_Y=None, xsum=0, xconst=0, ysum=0
             Xvark = Xvar
             Yvark = Yvar
         else:
-            Xvark = Xvar[:, column_selection].copy()
-            Yvark = Yvar[:, column_selection].copy()
+            Xvark = Xvar[:, column_selection]
+            Yvark = Yvar[:, column_selection]
         Cxxyy = _M2_dense(Xvar, Xvark, weights=weights, diag_only=diag_only) \
                 + _M2_dense(Yvar, Yvark, weights=weights, diag_only=diag_only)
         Cxy = _M2_dense(Xvar, Yvark, weights=weights, diag_only=diag_only)
@@ -581,12 +581,12 @@ def _M2_symmetric(Xvar, Yvar, mask_X=None, mask_Y=None, xsum=0, xconst=0, ysum=0
                 ykvarsum = yvarsum
                 ykconst = yconst
             else:
-                Xvark = Xvar[:, _filter_variable_indices(mask_X, column_selection)].copy()
+                Xvark = Xvar[:, _filter_variable_indices(mask_X, column_selection)]
                 mask_Xk = mask_X[column_selection]
                 xksum = xsum[column_selection]
                 xkvarsum = xksum[mask_Xk]
                 xkconst = xconst[_filter_variable_indices(~mask_X, column_selection)]
-                Yvark = Yvar[:, _filter_variable_indices(mask_Y, column_selection)].copy()
+                Yvark = Yvar[:, _filter_variable_indices(mask_Y, column_selection)]
                 mask_Yk = mask_Y[column_selection]
                 yksum = ysum[column_selection]
                 ykvarsum = yksum[mask_Yk]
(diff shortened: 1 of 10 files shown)


pyemma/coordinates/api.py
```diff
@@ -1414,9 +1415,113 @@ def vamp(data=None, lag=10, dim=None, scaling=None, right=True, ncov_max=float('
     return res
 
 
+def nystroem_tica(data, lag, max_columns,
```
franknoe: Please rename to ''tica_nystroem'' (makes it easier to find)

franknoe (commented):
Good for me except for the minor name change indicated here. It would be good if @fabian-paul  or @fnueske could have a look too.

pyemma/coordinates/api.py
florianlitzinger: OK, done.

marscher:
Thanks for this major contribution! In general the code looks also good to me. Here are my 50 cents:

A lot of code has been changed, but unfortunately not tested, especially in the moments module (https://codecov.io/gh/markovmodel/PyEMMA/pull/1261/diff?src=pr&el=tree#diff-cHllbW1hL19leHQvdmFyaWF0aW9uYWwvZXN0aW1hdG9ycy9tb21lbnRzLnB5 - every line which is in a big green area has been added and if the bar on the left is red, the code has not been executed during the tests. So I doubt that this code is either needed, or you need to demonstrate the code is working by providing a test case (untested code is unfortunately most often broken code). So I guess all of the sparse
input things have not been tested.

In the test for the Nyström method you generate a pseudo sparse matrix, in which 10 out of 100 columns have a meaning. However the order of these columns is fixed, right? I would suggest to shuffle the columns with a meaning to make this test even harder.

A lot of code has been copied from the tica module (which codacy also detected). This indicates we should introduce a common base class for PCA, TICA, VAMP and now Nystroem, since they all share eg. the handling of dimensions. Lazy diagonalization is also copied all over the place. Duplicated code makes it harder to change things in the future. But I guess we need to change this, as there should be some discussion which functionality to share.

Just another minor thing, please use new-style objects for your classes (derive them from object).

pyemma/coordinates/api.py
```diff
@@ -1414,9 +1415,113 @@ def vamp(data=None, lag=10, dim=None, scaling=None, right=True, ncov_max=float('
...
 
+def tica_nystroem(data, lag, max_columns,
+                  dim=-1, var_cutoff=0.95, epsilon=1e-6,
+                  stride=1, skip=0, reversible=True, ncov_max=float('inf'),
+                  initial_columns=None, nsel=1, neig=None):
+    r""" Sparse sampling implementation [1]_ of time-lagged independent component analysis (TICA) [2]_, [3]_, [4]_.
+
+    Parameters
+    ----------
+    data : ndarray (T, d) or list of ndarray (T_i, d) or a reader created by
```
marscher: the docstring says data is optional, but it isn't. All of the other estimators in coordinates api support just the construction of the object without estimation.

pyemma/coordinates/transform/nystroem_tica.py
```diff
@@ -0,0 +1,921 @@
...
+        # check nsel
+        nsel = self._check_nsel()
+        if nsel is None:
+            return None
+        # go
+        return self._select(nsel, err)
+
+    def _select(self, nsel, err):
+        """ Override me to do selection """
+        pass
```
marscher: raise NotImplementedError here, so this fails, if a deriving class forgot to overload this.

pyemma/coordinates/transform/nystroem_tica.py
```diff
@@ -0,0 +1,921 @@
...
+
+
+class SelectionStrategyRandom(SelectionStrategy):
```
marscher: test me

pyemma/coordinates/transform/nystroem_tica.py
```diff
@@ -0,0 +1,921 @@
...
+        # do select
+        sel = []
+        while len(sel) < nsel:
+            i = np.random.choice(self._oasis_obj._n)
+            if not (i in sel or i in self._oasis_obj._columns):
+                sel.append(i)
+        return np.array(sel, dtype=int)
+
+
+class SelectionStrategyOasis(SelectionStrategy):
```
marscher, line 846: test me

pyemma/_ext/variational/estimators/moments.py
```diff
@@ -359,20 +359,47 @@ def _center(X, w, s, mask=None, const=None, inplace=True):
     return X, const
 
 
+def _filter_variable_indices(mask, column_selection):
```
marscher, line 362: this function is never called.

pyemma/_ext/variational/estimators/moments.py
```diff
@@ -521,14 +567,34 @@ def _M2_symmetric(Xvar, Yvar, mask_X=None, mask_Y=None, xsum=0, xconst=0, ysum=0
...
+                Xvark = Xvar
+                mask_Xk = mask_X
+                xkvarsum = xvarsum
+                xkconst = xconst
+                Yvark = Yvar
+                mask_Yk = mask_Y
+                ykvarsum = yvarsum
+                ykconst = yconst
+            else:
+                Xvark = Xvar[:, _filter_variable_indices(mask_X, column_selection)].copy()
```
marscher: why do you take a copy here? Is the selection modified in place?

pyemma/coordinates/tests/test_nystroem_tica.py
```diff
@@ -82,5 +90,79 @@ def test(self):
         assert np.all(np.abs(self.tica_obj.eigenvalues[:4] - t.eigenvalues[:4]) < 1e-3)
 
 
+class TestNystroemTICA_oASIS(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.X = np.ones((10000, 10))
+        cls.variable_columns = np.random.choice(10, 5, replace=False)
```
marscher, line 97: Thanks! But does the test also pass if you make the choice out of np.random.choice(cls.X.shape[1], 5, replace=False)

marscher:
Thanks for the bugfixes and additional tests. The Codacy issues are mostly about copy paste code, but I think we can restructure things after merging, since it requires some additional design discussions. The issue about the not initialized member variable is actually a false positive.

pyemma/coordinates/tests/test_nystroem_tica.py
florianlitzinger: How do you mean? (cls.X.shape[1] == 10)

marscher:
Indeed, makes no sense. Can you blow up the matrix to a higher dimension 
(eg. 100 or even more) and place 10 meaningful columns at random indices?

PR #1264, opened 2018-03-01
Histogram plot

thempel:
This multi-dimension histogram plot function was requested at the last pyemma workshop after I showed something similar in my presentation. It's nice to have but not very quantitative since histograms are scaled in order to be visible even for many dimensions. It produces graphs like this: 
multidim_hist

### DIFF
--- a/pyemma/plots/plots1d.py
+++ b/pyemma/plots/plots1d.py
@@ -35,7 +35,7 @@ def plot_feature_histograms(xyzall, feature_labels=None, ax=None, ylog=False, ou
         the axes to plot to. When set to None the default Axes object will be used.
     ylog : If True, plot logarithm of histogram values.
     n_bins : Number of bins the histogram uses.
-    outfile : If not None, saves plot to this file. 
+    outfile : If not None, saves plot to this file.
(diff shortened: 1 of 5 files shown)


franknoe:
Nice! 


pyemma/plots/plots1d.py
```diff
@@ -0,0 +1,107 @@
...
+    xyzall : (Concatenated list of) input features; np.array containing time series data to be plotted.
+    feature_labels : Labels of histogrammed features, optional, default = None.
+                  Either list of strings or pyemma Featurizer object.
+    ax : matplotlib Axes object, optional, default = None
+        the axes to plot to. When set to None the default Axes object will be used.
+    ylog : If True, plot logarithm of histogram values.
+    n_bins : Number of bins the histogram uses.
+    outfile : If not None, saves plot to this file.
+    ignore_dim_warning : Enable plotting for more than 50 dimensions (on your own risk).
+    **kwargs: Will be parsed to pyplot.fill_between when plotting the histograms.
```
marscher: passed to?

marscher:
Very nice figure :+1: Can you add this fancy new feature to the change log as well? Then we're good to merge.

marscher:
Thank you again! Good job!

PR #1267, opened 2018-03-02
[ML-HMSM] fix the setting of nstates (got overriden by MSM.nstates.fset)

### LINKED ISSUE #1266: timescales_hmsm with n_jobs > 1 not working
With the latest PyEMMA release 2.5.1, it seems that using multiprocessing with implied timescales estimation for HMMs is somewhat broken. The following minimal example reliably crashes:

```python
import pyemma
import numpy as np
dtrajs =[np.random.randint(0, 10, 1000) for _ in range(10)]
pyemma.msm.timescales_hmsm(dtrajs, 2, lags=[1, 2, 3], n_jobs=2)
```

It raises `RuntimeError: Estimation has failed at ALL lagtimes.` Works with `n_jobs=1` or release 2.4. Can anyone reproduce this?

marscher:
The ML-HMSM class overrides the meaning of nstates defined in
the base class (MSM) from being a model parameter to be an input
parameter for estimation. By defining our own setter in the estimator
we avoid that this input parameter gets overriden (eg. by setting self.P
to None).

Changed its_all_estimator tests to run in parallel as it is a stronger
test, because it implies serialization and estimation.

fixes #1266

### DIFF
--- a/pyemma/msm/estimators/maximum_likelihood_hmsm.py
+++ b/pyemma/msm/estimators/maximum_likelihood_hmsm.py
@@ -289,6 +289,17 @@ class MaximumLikelihoodHMSM(_Estimator, _HMSM):
         return self.lag
 
     @property
+    def nstates(self):
+        return self._nstates
+
+    @nstates.setter
+    def nstates(self, value):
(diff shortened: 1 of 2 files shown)


thempel:
Many thanks for that fast fix!

PR #1269, opened 2018-03-06
[vamp] use gesvd as lapack method to solve the svd problem.

marscher:
There had been some numerical issues with small epsilons, by using
the gesvd method, these examples work as well. The svd raised a not converged error,
when using the default numpy.linalg.svd.

thanks @clonker for his great help

https://github.com/ContinuumIO/anaconda-issues/issues/695

### DIFF
--- a/pyemma/coordinates/transform/vamp.py
+++ b/pyemma/coordinates/transform/vamp.py
@@ -275,13 +275,11 @@ class VAMPModel(Model, SerializableMixIn):
               singular value. Note that only the left singular functions
               induce a kinetic map.
         """
-
         L0, self._rank0 = spd_inv_sqrt(self.C00, epsilon=self.epsilon, return_rank=True)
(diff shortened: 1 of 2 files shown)


fabian-paul:
Hi guys,

I think the current way how I implemented the SVD of the half-weighted Koopman matrix is not ideal. I just copy-and-pasted from the TICA implementation without much thinking. Because of this the current implementation performs an SVD of a matrix that is much larger and much more rank-deficient that it should be.
Here is the alternative implementation that I would rather prefer. If you're ok, I'm going to make a PR with this new implementation.

screenshot from 2018-03-06 09-47-08

marscher:
Of course a more robust computation would be preferable. Thank you very much for looking into this again.

Where does A and B come from in your equations?

fabian-paul:
A, Sigma and B are just the result that you get by calling the SVD routine on W.

marscher:
Ah just overlooked that, thanks. Then it looks good to me, since you're operating in the low-rank space after the first decompositions.

PR #1274, opened 2018-03-08
more reliable SVD implementation for VAMP, part 2

fabian-paul:
Hi all, 
while correcting the SVD implementation, I found a couple of issues that I correct here.
There already existed code for the reliable SVD (spd_inv_split), I just didn't understand it. Now that I've understood, I have changed _diagonalize() again. The changes in this PR are purely organizational. The actual SVD algorithm is unchanged. (We still do the SVD if C01 in the small space.)


marscher:
You should replace np.testing.assert_raises with self.assertRaises(ExceptionType) then we do not need Nosetests as testing dependency.

PR #1281, opened 2018-03-23
Fix sidechain torstions feat describe selection

marscher:
@gph82 can you take a look please?


gph82:
LGTM, checked with  pyemma.version 2.5.1+65.gd42bad53. Correct labeling.

marscher:
I guess you have checked this with and without a selection?

gph82:
Of course! with () (and all residues are added up to all available chis) and then for just one residue

marscher:
Thank you very much for testing this!

PR #1283, opened 2018-03-26
Init bhmm bug

thempel:
1) When estimating a `BayesianHMSM` with `init_hmsm`, it is tested if the initial HMM was estimated on the same data set. This test was broken.
2) The instance of `self.hmm = bhmm.DiscreteHMM()`was added to `__serialize_fields`.


marscher:
need to white-list bhmm.hmm.discrete_hmm for unpickling.

marscher:
Thanks again!

PR #1291, opened 2018-04-18
[StreamingTransformerIterator] propagate chunksize not only to state,…

### LINKED ISSUE #1290: StreamingTransformerIterator does not propagate chunksize properly
For convenience we assign the IteratorState of underlying iterator to the StreamingTransformerIterator. This assumes that all iterator implementations directly use the state, but this does not work as expected for the FragmentedIterator. The FragmentedIterator encapsulates another iterator, which has its own state. The chunksize will not be propagated, if set on the StreamingTransformerIterator.

By overriding the chunksize setter in StreamingTransformerIterator like this, we can avoid this situation.
``` python

    @DataSourceIterator.chunksize.setter
    def chunksize(self, val):
        self.state.chunk = val
        self._it.chunksize = val
```

marscher:
… but also to the underlying iterator.

Fixes #1290

### DIFF
--- a/pyemma/coordinates/data/_base/iterable.py
+++ b/pyemma/coordinates/data/_base/iterable.py
@@ -249,21 +249,23 @@ class _LaggedIterator(object):
             frag = data_lagged[:min(self._it.chunksize - self._lag, len(data_lagged)), :]
             data = np.concatenate((self._overlap, frag[(self._actual_stride - self._lag)
                                                        % self._actual_stride::self._actual_stride]), axis=0)
+            data_len = len(data)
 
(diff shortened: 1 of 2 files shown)


marscher:
@clonker do you think it is safe to broaden the assertion in the LaggedIterator that the actual data is also larger than the overlap? Because of this bug the big chunk had exactly the same size as the overlap.

clonker (approved):
Looks good to me!
Since the overlap is supposed to be the intersection of the data and time-lagged data, i think it is safe to make that assertion.

PR #1306, opened 2018-05-24
plots.plot_markov_model minor fix

thempel:
`pyemma.plots.plot_markov_model(P, state_labels=None)` would raise a 
```python
TypeError: object of type 'NoneType' has no len()
``` 
because this case was not handled properly in the `plot_network()` function.


marscher:
Thanks a lot! LGTM

marscher (approved):

PR #1308, opened 2018-05-30
[hmsm] hide ImaginaryEigenvalueWarning, when computing decorrelation time

marscher:
because this model is only used internally and it is non-reversible,
we do not want to scare users with this warning.


pyemma/msm/estimators/maximum_likelihood_hmsm.py
```diff
@@ -180,14 +180,20 @@ def _estimate(self, dtrajs):
             # by default use lag as stride (=lag sampling), because we currently have no better theory for deciding
             # how many uncorrelated counts we can make
             self.stride = self.lag
-            # get a quick estimate from the spectral radius of the nonreversible
+            # get a quick estimate from the spectral radius of the non-reversible
```
cwehmeyer, line 183: **nonreversible** is correct, too.

cwehmeyer (approved):
Looks good to me. Thanks!

PR #1315, opened 2018-06-11
Coordinates code clean up and massive testing

marscher:


marscher:
The dcd tests currently can not pass because of the fixed stride handling in mdtraj. This will require a new mdtraj release, which is scheduled.

pyemma/coordinates/data/_base/datasource.py
```diff
@@ -246,9 +246,9 @@ def trajectory_length(self, itraj, stride=1, skip=0):
             selection = stride[stride[:, 0] == itraj][:, 0]
             return 0 if itraj not in selection else len(selection)
         else:
+            assert skip is not None
             skip = 0 if skip is None else skip
```
clonker: if skip is asserted to be `not None` this line is unnecessary

pyemma/coordinates/data/_base/datasource.py
```diff
@@ -289,8 +289,8 @@ def trajectory_lengths(self, stride=1, skip=0):
                                 for itraj in range(n)),
                                dtype=int, count=n)
         else:
-            return np.fromiter(((l - skip - 1) // stride + 1 for l in self._lengths),
-                               dtype=int, count=n)
+            return np.fromiter((max(0, ((l - skip - 1) // stride + 1)) for l in self._lengths),
```
clonker: can this make a call to `trajectory_length`? seems like code duplication otherwise

pyemma/coordinates/data/_base/datasource.py
```diff
@@ -727,10 +737,31 @@ def close(self):
         """ closes the reader"""
         raise NotImplementedError()
 
+    def _select_file_guard(f):
+        """
```
clonker: empty documentation?

pyemma/coordinates/data/_base/datasource.py
```diff
@@ -824,9 +862,26 @@ def _itraj(self, value):
         value : int
             The upcoming trajectory index.
         """
-        if value > self.state.ntraj:  # we never want to increase this value larger than ntraj.
-            raise StopIteration("out of files bound")
-        self.state.itraj = value
+        if value != self._selected_itraj:
+            self.state.itraj = value
+            # TODO: this side effect is unexpected.
```
clonker, line 864: i propose removing the setter and introducing a method, e.g., `ds._change_trajectory(itraj)` - then the side effect is not that unexpected anymore

pyemma/coordinates/data/_base/datasource.py
```diff
@@ -659,12 +664,15 @@ def __init__(self, data_source, skip=0, chunk=0, stride=1, return_trajindex=Fals
                                    ntraj=self.number_of_trajectories(),
                                    cols=cols)
         self.__init_stride(stride)
-        self._pos = 0
         self._last_chunk_in_traj = False
+        # the currently selected itraj, used to distinguish self._itraj in _select_file
+        # TODO: this is duplicated with self.state.current_trajindex or so?!
+        self._selected_itraj = -1
+        #TODO: replace with self._skip_unselected_trajs?
```
clonker: yes, I would replace that

pyemma/coordinates/data/_base/datasource.py
```diff
@@ -963,60 +1018,68 @@ def last_chunk_in_traj(self):
...
-                self._itraj += 1
-            self._select_file(self._itraj)
-        # we have to obtain the current index before invoking next_chunk (which increments itraj)
-        self.state.current_itraj = self._itraj
+    def __next__(self):
+        # the position is the previous advanced position
         self.state.pos = self.state.pos_adv
+
+        # increase itraj
+        self._skip_unselected_or_too_short_trajs()
```
clonker: this call can probably be removed since initially the too-short trajectories are skipped and we also skip post-hoc after every chunk yield

pyemma/coordinates/data/_base/datasource.py
```diff
@@ -963,60 +1018,68 @@ def last_chunk_in_traj(self):
...
+
+        self.state.pos_adv = self._t
+        if self.return_traj_index:
+            return self.state.current_itraj, X
         return X
 
+    next = __next__
+
+    @staticmethod
+    def __chunk_finite(data):
```
clonker: this could go into the data source and be then also used in the `get_output` finiteness check

pyemma/coordinates/data/_base/iterable.py
```diff
@@ -73,6 +73,7 @@ def default_chunksize(self):
         """
         if self._default_chunksize is None:
             try:
+                # TODO: if dimension is not yet fixed (eg tica var cutoff, use dim of data_producer.
```
clonker, line 76: )

pyemma/coordinates/data/_base/iterable.py
```diff
@@ -292,44 +294,80 @@ def __init__(self, it, it_lagged, return_trajindex):
...
+            it._select_file(sufficiently_long_traj_indices[0])
+            break
+
+        idx = (np.abs(np.array(sufficiently_long_traj_indices) - it._itraj)).argmin()
+        if idx + 1 < len(sufficiently_long_traj_indices):
+            next_itraj = sufficiently_long_traj_indices[idx + 1]
+            it._select_file(next_itraj)
+            assert it._itraj == next_itraj
+        else:
+            raise StopIteration('no trajectory long enough.')
```
clonker, line 371: we should think about what is raised when this condition is met in the constructor

clonker (commented):
code got a lot cleaner, IMO :+1:

PR #1317, opened 2018-06-16
Refactor plots2d

cwehmeyer:
This PR refactors the plots2d module:

- auxiliary functions allow to make custom plots easily
- new plotting functions (plot_density, plot_free_energy_new, plot_contour) with more control (e.g. color bars)
- old plotting functions retain signatures and behaviour but use new implementation


doc/source/CHANGELOG.rst
```diff
@@ -16,6 +16,8 @@ Changelog
 - msm: fixed serialization of BayesianHMSM, if initialized with a ML-HMSM. #1283
 - base: require at least tqdm >= 4.23, because of an API change. #1292, #1293
 - plots: fixed minor bug in plot_network (state_labels=None would not work). #1306
+- plots: refactored plots2d
```
marscher: this message is kind of uninformative to the user. I would state, that the gca behaviour has changed etc.

pyemma/plots/plots2d.py
```diff
@@ -121,23 +114,27 @@ def plot_free_energy(xall, yall, weights=None, ax=None, nbins=100, ncountours=10
...
         sample weights. By default all samples have the same weight
     ax : matplotlib Axes object, default = None
-        the axes to plot to. When set to None the default Axes object will be used.
+        the axes to plot to. When set to None the default Axes object
+        will be used.
     nbins : int, default=100
         number of histogram bins used in each dimension
     ncountours : int, default=100
         number of contours used
     offset : float, default=0.1
```
marscher: default is -1 now.

pyemma/plots/plots2d.py
```diff
@@ -156,33 +153,403 @@ def plot_free_energy(xall, yall, weights=None, ax=None, nbins=100, ncountours=10
     ax : Axes object containing the plot
 
     """
-    import matplotlib.pylab as _plt
+    #import matplotlib.pylab as _plt
```
marscher: delete me

pyemma/plots/plots2d.py
```diff
@@ -156,33 +153,403 @@ def plot_free_energy(xall, yall, weights=None, ax=None, nbins=100, ncountours=10
...
+    x, y, z = get_histogram(
+        xall, yall, nbins=nbins,weights=weights,
+        avoid_zero_count=avoid_zero_count)
+    return plot_map(
+        x, y, _to_density(z).T, ncontours=ncontours, ax=ax,
+        cmap=cmap, cbar=cbar, cax=cax, cbar_label=cbar_label,
+        logscale=logscale)
+    
+
+def plot_free_energy_new(
```
marscher: what is actually the difference to the old method? Can we decide to have only one or are they fundamentally different?
A short line in the docstring what this method does different would be nice.

marscher (approved):
Thank you very much for cleaning this up! Especially not using gca in cases where it is inappropriate is a big win.

pyemma/plots/plots2d.py
cwehmeyer: The new plotting functions always return `fig, ax, cbar` while the old functions return either `fig, ax` or just `ax`. Further, `plot_free_energy` has a parameter typo (`countour`) which has been fixed in the new function.

I would love to immediately replace everything with the new functions but this would slightly break the API.

marscher: We could have a legacy switch parameter to steer the old or new behaviour, but this would convolute the code a bit. I think we should deprecate the old function and promote the new one, shouldn't we?

cwehmeyer: Yes, indeed. I don't think functions would be too convoluted, though. Using a legacy switch is a good idea!

cwehmeyer: Do we need to conserve the sequence of named parameters in legacy mode?

marscher:
> 
> Do we need to conserve the sequence of named parameters in legacy mode?
> 

strictly speaking yes, because if somebody put in the parameters in the
previous order, you would break the intended usage of the parameter.

cwehmeyer:
@thempel please have a look at `test_feature_histograms_nowarning` in `TestPlots1d`.

thempel:
Got it. This test was too unspecific and pretty much redundant (checked if no warning is raised for this particular function). Also, it did not show the raised warning. I removed the test and put some more meaningful ones there instead. 
Anyway, probably the failure was caused by this warning. Don't know if it's important, but just for the record.

RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).     max_open_warning, RuntimeWarning)

thempel:
For `plot_density()`, I get different behavior if using logscale or not. Logscale produces a discrete colormap which is otherwise continuous. Was that intended?

thempel:
@cwehmeyer we discussed that it is potentially useful to have access to colormap or mappable. Since I found it a bit inconsistent to return either the colormap or the mappable, depending on the kwargs, they are now written into a dictionary and returned together. Otherwise, larger customized plots might break down if somebody chooses to alter some of the plot options. I'm not completely happy with it though; if you have a better idea feel free to implement it.

thempel (approved):

thempel:
Thx @cwehmeyer. Can you add a line about the kwargs (i.e. what function they are passed to) to the docstrings?

cwehmeyer:
@thempel: done.

thempel (approved):

PR #1323, opened 2018-06-26
HMM submodel fix

thempel:
Calling the `submodel()` function of a (Bayesian) HMM would not only return a sub-model but indeed overwrite the HMM instance directly. This makes sense when used within `_estimate()`, but not if a user wants to have an actual submodel. The PR submodels and returns a copy of `self` in these cases.


marscher (approved):
Thanks a lot. I like the inplace parameter much more than the internal state hack we added yesterday!

PR #1326, opened 2018-06-27
[msm tests] hard-code submodel test dtraj

thempel:
new HMM / BHMM submodel test fails stochastically, fixed dtrajs.


marscher (approved):
this is the weakly connected dtraj, right?

thempel:
Yes, one that is sufficiently strongly connected to make it work on the full set and weakly enough to be taken apart by `submodel_largest(mincount_connectivity=5)`

Issue #1334, opened 2018-07-04
[plots] zorder parameter


thempel:
For overplotting `pyemma.plots.plot_contour()` with `pyemma.plots.plot_flux()` or anything network plotting related, I keep getting unreadable figures due to arrows being hidden behind the contour. That spoils the pyemma_tutorials plots I was making. @cwehmeyer do you think we can fix this? `matplotlib.contourf()` does not naturally support `zorder` kwargs, if I'm not mistaken. 
![asdf](https://user-images.githubusercontent.com/12544861/42285442-ef9be1a6-7faf-11e8-89ec-759f7980d79d.png)

cwehmeyer:
You can use the `alpha` parameter in `pyemma.plots.plot_contour()`.

thempel:
-.-

PR #1336, opened 2018-07-05
Allow zorder in plots2d

cwehmeyer:
With this, we can pass the `zorder` parameter to the `pyemma.plots.plots2d` functions as requested in #1334.


cwehmeyer:
The CI errors are unrelated.

thempel:
Reverting ill-defined circle fix because of possible implications for existing plots.

thempel (approved):
Thanks for working this out!

PR #1337, opened 2018-07-05
[appyevor] use miniconda 3.6 as base and rely on CONDA_PY

marscher:
CONDA_PY selects the actual python version used in the build.

This should fix the build on 32 bit and python 3.5


marscher:
note, that conda-build does not respect CONDA_NPY and --numpy to set the numpy runtime version.

PR #1339, opened 2018-07-09
[kmeans] only show progress for creating data array, if src is not in memory.

marscher:
If we show a pg for in memory data, we will see like one iteration is made then the process is already done, because it is soo quick. This has lead to ugly outputs.


marscher:
failures are unrelated.

PR #1343, opened 2018-07-26
Fix connected sets for Bayesian MSMs

marscher:
These sets were not computed on the same count matrix as the one used to compute the active set.
    
This had lead to differences between the largest connected set and
the shape of P and pi. When the ergodic cutoff kicked out low populated
states in during count matrix estimation, the connected sets were estimated
on a wrong measure, because it used the full count matrix instead of the truncated one.

### DIFF
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -1,6 +1,24 @@
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


thempel (approved):
Maybe add a unit-test that compares active set to largest connected set?

Issue #1349, opened 2018-08-24
koopman bug and fix


euhruska:
Found a bug in the koopman implementation by analysing https://github.com/radical-collaboration/extasy-grlsd/issues/77 together with @fnueske and a suggestion how to solve it.
Take line
https://github.com/markovmodel/PyEMMA/blob/1a280481ed6cbbfc3fbceaacfa22136e26983678/pyemma/_ext/variational/solvers/direct.py#L82
 with ```s = _np.array([ 1.90963508, 1.72242421, 1.18440072, -1.16845181])```
leads to an m=4 in line https://github.com/markovmodel/PyEMMA/blob/1a280481ed6cbbfc3fbceaacfa22136e26983678/pyemma/_ext/variational/solvers/direct.py#L89
, while m=3 is expected (as for 3 positive eigenvalues, removing later the 4th)

Suggestion: increase the value 1e-16 to 1e-14 in line
https://github.com/markovmodel/PyEMMA/blob/1a280481ed6cbbfc3fbceaacfa22136e26983678/pyemma/_ext/variational/solvers/direct.py#L84
This will lead to m=3, but not sure if that's the best fix as this might depend on the individual machine precision of the user

System: pyemma 2.5.4, python 2.7.14, numpy

marscher:
solved, was a bug, caused by machine precision

_Originally posted by @euhruska in https://github.com/radical-collaboration/extasy-grlsd/issues/77#issuecomment-421409376_

PR #1350, opened 2018-08-24
fix koopman bug

euhruska:
described in https://github.com/markovmodel/PyEMMA/issues/1349


marscher:
Could you please turn your failing example into a test case?

PR #1366, opened 2018-09-26
Tica model analog to vamp model

### LINKED ISSUE #1075: Simplify TICA Parameters
``tica``'s init method has four parameters that will determine the output dimension, and that have accumulated over time as different methods were added (especially kinetic maps and commute maps):
  * dim=-1
  * var_cutoff=0.95
  * kinetic_map=True
  * commute_map=False

Now that's a bit confusing, because these parameters are interdependent, and also the current default behavior is not the most sensible. By default, I will get a scaling by TICA eigenvalues (to a kinetic map at the selected lag time), and the output dimension will be selected such that the kinetic variance adds up to 95%. If I instead want to select a fixed output dimension, I would have, in principle to set two values:
`dim=10, var_cutoff=1.0`. To avoid that, we check if the var_cutoff was set away from the default, i.e. setting `dim=10, var_cutoff=0.95` will end up in 10 dimensions, even if the variance is 99%. That's confusing. Moreover, the current situation allows illegal or inconsistent settings such as `kinetic_map=True, commute_map=True`.

Looking at the sklearn impl of PCA, which has to address the same situation, here's a suggestion how to solve it. We only keep the following two parameters to control the output dimension and scaling:

```python
   """
        dim : float or int
            Number of dimensions to keep:
            * if dim is not set all available ranks are kept::
                n_components == min(n_samples, n_features)
            * if dim is an integer >= 1, this number specifies the number
              of dimensions to keep. By default this will use the kinetic
              variance unless scaling=`commute map` is selected.
            * if dim is a float with ``0 < dim < 1``, select the number
              of dimensions such that the amount of kinetic variance 
              that needs to be explained is greater than the percentage 
              specified by dim.
        scaling : None or string
            Scaling to be applied to the TICA modes upon transformation
            * None: no scaling will be applied, variance along the mode is 1
            * 'kinetic map' or 'km': modes are scaled by eigenvalue
            * 'commute map' or 'cm': modes are scaled by :math:`sqrt(t_i/2)`, 
              where :math:`t_i=-lag/|lambda_i|` is the relaxation time computed 
              from the eigenvalue :math:`lambda_i`.
   """
```

That way, `dim` is the only parameter that determines the output dimension, and it can be set in different ways to do that. `scaling` is the only parameter that determines the output of `transform(X)` or `get_output()`, it could get more complex (i.e. additional paramters in the string, or accepting a dict),
if we wanted to encode more, such as the function use to penalize small eigenvalues.

Of course we would need to introduce this behavior "smoothly", i.e. without killing the other parameters immediately, but we could enable the new parameter behavior first, deprecate the old parameters and remove them in a later version.

We would do the same changes in `pca`, although the situation there is a bit simpler.

What do you think? Is it worth making this change in PyEMMA 2.*, or does it add confusion?

marscher:
In this PR we moved the handling the dimension and diagonalization from TICA to its model. This turned out to be good solution for the new VAMP estimator/model pair and therefore we should use it from now on for TICA and its children.

In VAMP the dim parameter can either take a fixed dimension (integer), a float for variance cutoff, or None to preserve all available ranks.
To handle optional scaling of the singular/eigenvectors it has a 'scaling' parameter taking strings like 'kinetic_map' etc.

This approach should fix #1075, since it reduces the amount of input parameters and invalid combinations.

The default behavior is not changed, it still uses kinetic map scaling and preservation of 95% of kinetic variance.

The changes have also been applied for NystroemTICA. This avoids hard to maintain copy paste code, as this implementation copied over almost everything from TICA except for the low-rank, column selection section.

Already serialized TICA objects are future proof as well, as these are transformed during loading to fit into the new class layout.


marscher:
@fabian-paul I'd be delighted, if you could have a look at the changes I've made to TICA, especially the newly introduced rank handling in the dimension reduction. If you think this PR is too big, I could try to tidy/split it up.
Thanks in advance!

marscher:
:*(

PR #1376, opened 2018-11-07
[plots] fixed vmin, vmax kwarg pass to subsequent function call

thempel:
Keyword arguments `vmin` and `vmax` were processed but not passed from `pyemma.plots.plot_contour()` to its internal call of `pyemma.plots.plot_map()`.


marscher:
LGTM

PR #1377, opened 2018-11-07
[msm/its] added 'only_timescales' option

marscher:
Avoid blowing up memory, if one estimates ITS with BMSM and lots of
states and or samples.


thempel (approved):
lgtm

marscher:
todo: add plotting test.

PR #1378, opened 2018-11-22
[coordinates/source] do not probe h5 file for mdtraj.HDF5TrajectoryFile

marscher:
The exception triggering and catching was too broad/unspecific to be
triggered in certain cases.

Now we just check, if the user wants to read MD data by checking if
either a topology or a featurizer has been given.


clonker (approved):
Looks good to me!

marscher:
Thanks for reviewing!

PR #1380, opened 2018-12-05
fix docstring format

sunhwan:
simple doctoring format fix


marscher:
Thank you! I wonder how many of these mistakes we have in our documentation.

PR #1382, opened 2018-12-06
Fix len for ascii trajs

marscher:


marscher:
fyi @thempel

marscher:
you're welcome!

PR #1388, opened 2019-01-30
added random seed option for double well

brookehus:
I decided it was easier to add a parameter to `init` but let me know if you prefer something different. What kind of tests do you want for this?

- [ ] Make sure to include one or more tests for your change
- [x] Add yourself to `AUTHORS`
- [x] Add a new entry to the `doc/source/CHANGELOG` (choose any open position to avoid merge conflicts with other PRs).
      Decide whether your change is a fix or a new feature.

### DIFF
--- a/pyemma/datasets/double_well_thermo.py
+++ b/pyemma/datasets/double_well_thermo.py
@@ -25,7 +25,7 @@ __all__ = [
 
 class DoubleWellSampler(object):
     '''Continuous multi-ensemble MCMC process in an asymmetric double well potential'''
-    def __init__(self):
+    def __init__(self, random_seed=None):
         self.xmin = -1.8
         self.xmid = 0.127
         self.xmax = 1.7
@@ -38,6 +38,12 @@ class DoubleWellSampler(object):
         self.pi = _np.exp(-epot)
         self.pi[:] = self.pi / self.pi.sum()
         self.f = -_np.log(self.pi)
+
+        if random_seed is not None:
+            self._rng = _np.random.RandomState(random_seed)
+        else:
+            self._rng = _np.random
+        self.random_seed = random_seed
 
     @property
     def centers(self):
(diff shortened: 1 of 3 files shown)


brookehus:
cc @thempel

pyemma/datasets/double_well_thermo.py
```diff
@@ -39,6 +39,12 @@ def __init__(self):
         self.pi[:] = self.pi / self.pi.sum()
         self.f = -_np.log(self.pi)
 
+        if random_seed is not None:
+            self._rng = _np.random.RandomState(random_seed)
```
marscher, line 43: does this overwrite the state for all numpy users? If that is the case, is there a way to circumvent reseeding others random state?

brookehus: No, I don't think so. I tested:

```
outer_rng = np.random.RandomState(100)
outer_rng.randn(1)
dw_sampler = DoubleWellSampler(random_seed=1)
outer_rng.randn(1)
p1 = dw_sampler.sample(length=10)['trajs'][0]
print(np.random.randn(1))
print(outer_rng.randn(1))
```

And the `outer_rng` is unaffected by the `random_seed` argument to `DoubleWellSampler` nor its execution, and the `np.random` call is different on different runs with the same `outer_rng` and `random_seed`.

brookehus: As far as I know the random state is only for calls that come from `_rng` in place of `_np.random`.

brookehus: One consideration though is that if the same `DoubleWellSampler` object were used to run `sample` more than once without a new instance of the `Sampler`, then each result would be different.

marscher: Thanks for testing this. Why would this be an issue in the latter case? I would expect it to be different.

brookehus: Yeah, same. Not an issue.

marscher:
Should I wait for a test or not? This seems like a minor change, which is not really exposed to user api.

brookehus:
If there should be a test let me know what I should test. I don't mind either way.

marscher:
LGTM

PR #1394, opened 2019-02-04
Make left singular functions the default in VAMP

fabian-paul:
Make left singular functions the default in VAMP. Also added kinetic map scaling of the right singular functions if the user wants it.


marscher:
Thank you!

PR #1395, opened 2019-02-05
Python3 only

marscher:
Finally removed Python2 compatibility :relieved:


marscher:
Appveyor crashes one test randomly and this used to happen before this change.

PR #1440, opened 2019-11-19
[util] improve performance of type check

### LINKED ISSUE #1439: MSM estimation 10x slower when dtraj is a list vs array
Super random - I noticed there was a large speed difference in another package depending on whether I fed dtrajs as a list or a numpy array -- thought I'd check how PyEMMA behaves, turns out lists are over 10x slower:

```python
traj = [np.random.randint(100) for i in range(1000000)]
traj_array = np.array(traj)

timeit -r 5 -n 5 pyemma.msm.estimate_markov_model(traj, 1)
1.07 s ± 16.8 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)

timeit -r 5 -n 5 pyemma.msm.estimate_markov_model(traj_array, 1)
80.4 ms ± 1.65 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)
```

clonker:
Small change to improve performance of check whether is something a list of int by a factor of roughly 20:
Old implementation on list of `1000000` random ints:
```
861 ms ± 9.3 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
New implementation on list of `1000000` random ints:
```
48.3 ms ± 932 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Fixes #1439

### DIFF
--- a/pyemma/util/types.py
+++ b/pyemma/util/types.py
@@ -25,6 +25,7 @@ import numpy as np
 import scipy.sparse as scisp
 import numbers
 import collections
+import itertools
 
 
 # ======================================================================================================================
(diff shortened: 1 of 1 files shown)


marscher:
That's brilliant, did you come up with the groupby type idea? :+1:

marscher:
the failed MSM tests are due to a change of version in openblas, I have reported it on Conda-forge (since it is actually on an increase in build number, not source code version) https://github.com/conda-forge/blas-feedstock/issues/40

clonker:
What can I say, cosmic radiation! :man_shrugging:

marscher:
The speed penalty of the old impl has to arise from the boolean and operations per element, not the type check. Since groupby reduces the boolean operations to the amount of types in the list, it is blazingly fast :zap:

PR #1449, opened 2020-01-29
Fixes #1443

### LINKED ISSUE #1443: mfpt questions
Hi,

 I'm using tram estimator from PyEMMA (2.5.6) and computing mfpt with the msmobj which obtained from tram calculations. I found some problems and hope you can help me figure out.

A.) Here is an msmobj I got from tram calculation.
array(ThermoMSM(P=array([[9.99347e-01, 3.06864e-09, ..., 2.63313e-08, 1.40791e-04],
       [1.53822e-03, 7.30838e-01, ..., 0.00000e+00, 0.00000e+00],
       ...,
       [8.14649e-03, 0.00000e+00, ..., 7.59904e-01, 2.15520e-03],
       [6.94215e-01, 0.00000e+00, ..., 3.43485e-05, 2.90880e-01]]),
     active_set=array([ 0, 19, 26, 41, 48, 50, 60, 64, 65, 73, 76, 77, 78, 91],
      dtype=int32),
     dt_model=[TimeUnit 200.0 step], neig=14, nstates_full=97,
     pi=array([9.96825e-01, 0.00000e+00, ..., 0.00000e+00, 7.69390e-07]),
     reversible=True), dtype=object) 

When I want compute the mfpt from state [19, 26] to state [0], I use this command msm.mfpt([19, 26], [0]). I got an error: AssertionError: Chosen set contains states that are not included in the active set.
So that means, I should use msm.mfpt([1, 2], [0]). But I got an another error: RuntimeWarning: invalid value encountered in true_divide muX = nuX / np.sum(nuX).
Do you know what's the problem for it?

B.) Next, I 'd like use the same msmobj to do tpt calculation.
I used the below commands:
from pyemma import msm  
msm.tpt(mm, [2, 3], [0])  ### here, mm is the msmobj from tram result
I can't do it because there's an error: ValueError: operands could not be broadcast together with shapes (97,1) (14,14). 
I guess the reason is the shape of mm.P and mm.active_set are not equal the shape of mm.pi.
Do you have some suggestions?

C.) Last, I found two ways to compute mfpt, are they same things?
1.)  from pyemma import msm  
      msm.tpt(mm, A, B)  ### here, mm is the msmobj from tram result
2.) mm.mfpt(A,  B)   ### here, mm is the msmobj from tram result

When A  and B are both one state, the results (from a test) are 114531.30349866912 (from (1)) and 114531.30440262843 (from (2)). They're not exactly same?
When A are a few states, the results are 741.4762666303764 (from (1)) and 15980.518823519338 (from (2)). They're so different?

Thanks for any help!!

fabian-paul:

### DIFF
--- a/pyemma/thermo/estimators/DTRAM_estimator.py
+++ b/pyemma/thermo/estimators/DTRAM_estimator.py
@@ -274,9 +274,10 @@ class DTRAM(_Estimator, _MEMM, ThermoBase):
 
         models = []
         for i, (msm, acs) in enumerate(zip(fmsms, active_sets)):
+            pi_acs = _np.exp(self.therm_energies[i] - self.bias_energies[i, :] - self.conf_energies)[acs]
+            pi_acs = pi_acs / pi_acs.sum()
             models.append(_ThermoMSM(
-                msm, self.active_set[acs], self.nstates_full,
-                pi=_np.exp(self.therm_energies[i] - self.bias_energies[i, :] - self.conf_energies),
+                msm, self.active_set[acs], self.nstates_full, pi=pi_acs,
                 dt_model=self.timestep_traj.get_scaled(self.lag)))
 
         # set model parameters to self
--- a/pyemma/thermo/estimators/TRAM_estimator.py
+++ b/pyemma/thermo/estimators/TRAM_estimator.py
@@ -437,9 +437,10 @@ class TRAM(_Estimator, _MEMM, ThermoBase):
 
         models = []
         for i, (msm, acs) in enumerate(zip(fmsms, active_sets)):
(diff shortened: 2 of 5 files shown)


franknoe:
Looks good to me superficially - but I have not followed the state discussion in detail.

fabian-paul:
I'm a bit disconcerted by the fact that we have not spotted to problem before.

clonker:
Looks good to me, per model definition `pi` only acts on the active set so it also has to be restricted if the energies live on the full state space. Only thing that might be worth adding is a small test case in which a state gets dropped and we check that the shape of `pi` matches the number of active states.
Otherwise I am also ok with merging this @fabian-paul @franknoe .

fabian-paul:
Here are tests.

clonker:
Cool thanks, will merge once CI gives green light!

PR #1453, opened 2020-03-18
add python 3.8 to the build matrices

clonker:


clonker:
merging this now even though circleci has some problem, python 3.8 is not possible yet because some dependencies are not updated and cause dependency conflicts. the only thing this PR changes is the updated URLs for miniconda from continuum.io to anaconda.org

PR #1526, opened 2021-12-22
Only pass copy of trajectory list to mdtraj.load and (CI) update macos image

clonker:


clonker:
the failing job didn't actually fail but just timed out in the last step, merging this

Issue #1541, opened 2022-02-18
Issue loading trajectory with pyemma.coordinates.source when topology file changes


AnjaConev:
Hello, 

Thanks for this amazing package - it has been very fun to work with and very reliable!

I have encountered a strange issue when loading the trajectory with pyemma.coordinates.source.
(PyEmma version = 2.5.7 on Debian GNU/Linux 9)

Here is the example of what is happening:

I have a trajectory and a reference PDB file (X.xtc and X.pdb). They both have 10 atoms.
I load the trajectory for the first time with:

`pyemma.coordinates.source("./X.xtc", top="./X.pdb")` 

Everything works as expected.

Later in my workflow I have new files with the same names: X.xtc, X.pdb but they now refer to a new trajectory with 18 atoms.
I make the same call to load this new trajectory:

`pyemma.coordinates.source("./X.xtc", top="./X.pdb")` 

It now fails with the error:
_ValueError: xyz must be shape (Any, 10, 3). You supplied  (1, 18, 3)_

If I reload the package and call source again with the new files it works.

It seems like the pdb file name gets cashed somewhere and pyemma.coordinates.source still thinks that we are dealing with the old X.pdb.
I was able to work around this issue without reloading the package by running:
`
mdtraj_top = mdtraj.load("./X.pdb").topology
`
`
pyemma.coordinates.source("./X.xtc", top= mdtraj_top)
` 

The code for reproducing this example and the corresponding files are in example.zip
[example.zip](https://github.com/markovmodel/PyEMMA/files/8100468/example.zip)

I just wanted to report this as it might be an unexpected behavior and I was stuck for a while on this issue.

My python environment: [pip_list.txt](https://github.com/markovmodel/PyEMMA/files/8100036/pip_list.txt)

Full error message:
![error_message](https://user-images.githubusercontent.com/7367905/154751297-3887b58c-b872-441f-ae20-d95c697bac07.PNG)

thempel:
Hi, I can reproduce this error locally. I believe it's a problem with the topology cache. (This cache is there because it accelerates loading large sets of trajectories with the same topology.) As far as I see, it is implemented using an LRU-cache here: https://github.com/markovmodel/PyEMMA/blob/8c2bc84dc1f0230202f368b5799908c16ebd611b/pyemma/coordinates/util/patches.py#L40-L52

[...]

PR #1545, opened 2022-03-01
Improved topology LRU cache

clonker:
- LRU cache now depends on hash over the first MB of data as well as last modified and creation dates.
- began enabling `pathlib.Path` objects for source/load

fixes issue #1541


thempel (approved):
looks good to me

PR #1550, opened 2022-03-11
Remove bhmm as dependency

clonker:


pyemma/msm/estimators/maximum_likelihood_hmsm.py
```diff
@@ -195,7 +196,8 @@ def _estimate(self, dtrajs):
                 self.stride = int(min(self.lag, 2*corrtime))
 
         # LAG AND STRIDE DATA
-        dtrajs_lagged_strided = bhmm.lag_observations(dtrajs, self.lag, stride=self.stride)
+        from deeptime.markov import compute_dtrajs_effective
+        dtrajs_lagged_strided = compute_dtrajs_effective(dtrajs, self.lag, n_states=-1, stride=self.stride)
```
thempel, line 200: does this function do the same? `bhmm.lag_observations` is just a plain lag-stride thing, I don't know what effective does here?

pyemma/msm/estimators/maximum_likelihood_hmsm.py
```diff
@@ -204,54 +206,56 @@ def _estimate(self, dtrajs):
...
-        # run
-        hmm_est.fit()
-        # package in discrete HMM
-        self.hmm = bhmm.DiscreteHMM(hmm_est.hmm)
+        from threadpoolctl import threadpool_limits
+        with threadpool_limits(limits=1):
+            if self.msm_init == 'largest-strong':
+                hmm_init = init.discrete.metastable_from_data(dtrajs, n_hidden_states=self.nstates, lagtime=self.lag,
+                                                              stride=self.stride, mode='largest-regularized',
+                                                              reversible=self.reversible, stationary=True,
```
thempel: `stationary` flag defaults to `False` in the constructor - should be `self.stationary` or remove flag I suppose

thempel: Or not because that's just the initial HMM?

pyemma/msm/estimators/maximum_likelihood_hmsm.py
clonker: it does the same but saves me from calling a private function (effective striding is only performed if `stride=='effective'`)

pyemma/msm/estimators/maximum_likelihood_hmsm.py
clonker: This is just reproducing the old behavior -- whether stationary is true or false should optimally be up to the user but the pyemma api currently doesn't allow that

pyemma/msm/estimators/bayesian_hmsm.py
```diff
@@ -18,7 +18,7 @@
 
 
 import numpy as _np
-
+from deeptime.markov._util import lag_observations
```
thempel: Use `compute_dtrajs_effective` instead?

clonker: I should... :smile: let me change that

clonker:
merging this now, going to follow up with another PR to resolve hanging CI

PR #1565, opened 2022-05-25
Use spawn() for multiprocessing instead of fork()

sents:
This PR changes the multiprocessing method used for the pool in `estimate_param_scan`.

### The issue
The reason for this PR is the following issue:
When using pyemma with n_jobs>1 and OpenBLAS I observed the worker processes getting stuck in lapack calls,
idling forever.  When using netlib blas or n_jobs=1 this problem did not occur.

This is likely due to OpenMP not supporting use in the main process AND forked processes, as explained [here](https://bisqwit.iki.fi/story/howto/openmp/).

### The fix
Setting the multiprocess start method to `spawn` for the pool fixes this issue.

### This is a breaking change
Using spawn requires guarding the main process with
```python
if __name__ == "__main__":
    #code 
```
For scripts this is best practice, but this PR would still break scripts which do not do this.

Common interactive contexts such as IPython or jupyter-notebook do this automatically and do not
require an explicit `if __name__ == "__main__":` statement.

### Checklist
- [ ] Make sure to include one or more tests for your change
- [x] Add yourself to `AUTHORS`
- [ ] Add a new entry to the `doc/source/CHANGELOG` (choose any open position to avoid merge conflicts with other PRs).
      Decide whether your change is a fix or a new feature.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,3 +27,4 @@ Brooke Husic
 Tim Hempel
 Sander Roet
 Sebastian Falkner
+Finn Krein
--- a/doc/source/CHANGELOG.rst
+++ b/doc/source/CHANGELOG.rst
@@ -19,7 +19,8 @@ Changelog
 - Use :code:`int` instead of :code:`numpy.int` to solve :code:`numpy` 1.20
   DeprecationWarning :pr:`1504`
 - Umbrella sampling: fix periodic width initialization :pr:`1522`
-- Solve 2 code:`SyntaxWarnings: "is" with a literal.` in utils/reader_utils.py :pr:`1530` 
+- Solve 2 code:`SyntaxWarnings: "is" with a literal.` in utils/reader_utils.py :pr:`1530`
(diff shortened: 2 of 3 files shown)


clonker:
Hi @sents, thank you very much for your contribution! I would agree that spawn is preferable here, especially since it causes the program to hang. However to limit breakage of other people's codes, perhaps this should only performed on linux platforms (you can check the `platform.system()` for that). MacOS and windows are - per your linked article and since we compile with clang and MSVC - not affected.

sents:
I've looked into [multiprocess.context](https://github.com/uqfoundation/multiprocess/blob/master/py3.9/multiprocess/context.py) to see what happens on the other platforms.
The default context for `sys.platform == "win32" ` is spawn already. For `sys.platform == "darwin"` the default context is still fork, although this is listed as a [bug](https://github.com/uqfoundation/multiprocess/issues/65) in the issue tracker.

As this is not in `multiprocess` yet I'll add check for only linux. The default in multiprocess will probably change to spawn for macOS at some point.

clonker:
Thanks! I'll merge once CI gives green light.

PR #1574, opened 2022-09-08
Fix tests, use deeptime double well discrete data.

clonker:


clonker:
Needs new deeptime release to be on conda-forge

Issue #1578, opened 2022-10-22
Possibility of specifying the tolerance for the MSM class


weitse-hsu:
Hi, 
I am a beginner using PyEMMA and I was using PyEMMA to analyze a transition matrix directly parsed from a GROMACS expanded ensemble simulation. When instantiating the class `MSM` in `pyemma.msm` with the transition matrix (which was a numpy array), I got the following error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/Wei-TseHsu/gmxapi_21.4/lib/python3.9/site-packages/pyemma/msm/models/msm.py", line 96, in __init__
    self.set_model_params(P=P, pi=pi, reversible=reversible, dt_model=dt_model, neig=neig)
  File "/Users/Wei-TseHsu/gmxapi_21.4/lib/python3.9/site-packages/pyemma/msm/models/msm.py", line 148, in set_model_params
    self.update_model_params(P=P)
  File "/Users/Wei-TseHsu/gmxapi_21.4/lib/python3.9/site-packages/pyemma/_base/model.py", line 79, in update_model_params
    setattr(self, key, value)  # set parameter for the first time.
  File "/Users/Wei-TseHsu/gmxapi_21.4/lib/python3.9/site-packages/pyemma/msm/models/msm.py", line 186, in P
    raise ValueError('T is not a transition matrix.')
```

To my understanding, the issue occurred because the sum of each row in my input transition matrix was very close but not exactly 1. Specifically, I've tried to use the function `is_transition_matrix` in `deeptime.markov.tools.analysis` with varying tolerances (specified by `tol`) and `is_transition_matrix(A, tol=1e-7)` returned true. However, when `is_transition_matrix` was used during the instantiation of `MSM`, the tolerance was specified as 1e-8, so the error occurred. (See [Line 185 in `msm.py`](https://github.com/markovmodel/PyEMMA/blob/devel/pyemma/msm/models/msm.py#L185).) While I temporarily solved the problem by installing the package from the source with the developer option and modifying the source code, I'm wondering if it's worth adding an optional parameter in `MSM` so that the user can specify the tolerance for telling if a matrix is a transition matrix. 

By the way, I was using PyEMMA 2.5.12 on MacOS Catalina 10.15.7. A text file with the output of `pip list` has been attached: [pip_list.txt](https://github.com/markovmodel/PyEMMA/files/9843763/pip_list.txt)

weitse-hsu:
[...]

PR #1579, opened 2022-10-25
Added an optional argument for specifying the tolerance in the class MSM

weitse-hsu:
This PR is aimed to resolve issue #1578.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,3 +28,4 @@ Tim Hempel
 Sander Roet
 Sebastian Falkner
 Finn Krein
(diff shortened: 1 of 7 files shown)


clonker:
Thank you! Would you mind moving the new keyword argument to the end of the list? That way this is not a breaking change if people were using the arguments based on their position not with the keyword specifier. I will have to make a deeptime release sometime soon in order for the changes to propagate, I'll see to that next week. :slightly_smiling_face:

weitse-hsu:
That makes sense! I have just made the changes. Thanks!

PR #1592, opened 2023-04-03
Substitute uses of numpy.bool with numpy.bool_

evilsetg:
Substitute uses of numpy.bool with numpy.bool_
Since NumPy 1.20 the usage of `numpy.bool` is [deprecated](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations). As a replacement `numpy.bool_` can be used.


clonker:
Thanks @evilsetg, I think I have to do some householding with numpy soon to get the test pipeline up and running again. Merging!:)

